PR #294, opened 2026-05-03
feat(numeric): land Slice D `as` cast LLVM lowering matrix

mcereal:
## Summary

Lands the deferred numeric LLVM cast lowering matrix on top of PR #289's parser-only Slice D. The previous attempt at this PR (commit ffd2620) was reverted because it bundled a `dominant_type` tightening that broke the test suite — see the commentary in commit 985be98 for the full diagnosis.

## What ships here

- **Cast matrix in `lower_cast_expression`** (`compiler/src/llvm/expression_lowering/native/core_literals_lowering.sfn`): emits `sitofp` / `fptosi` / `sext` / `zext` (i1 source) / `trunc` / `fpext` / `fptrunc` per the source/target pair, plus a hard rejection for `as bool` (fix-it points at `x != 0`).
- **Shared LLVM-type predicates** in `core_operands.sfn`: `is_integer_llvm_type`, `is_float_llvm_type`, `integer_llvm_width`. Reusable for future numeric work.
- **Tests**: extended `compiler/tests/unit/numeric_cast_test.sfn` (parser/native-IR shape pinning) + new `compiler/tests/e2e/test_numeric_cast.sh` (8 LLVM-shape pinning cases — sitofp / fptosi / sext / zext / trunc / fpext / fptrunc).
- **Self-host clean.** Seed stays at `0.5.10-alpha.8`, no seed bump needed.

## Smoke output

```text
fn convert(x: int)   -> float { return x as float; }  →  sitofp i64 %x to double
fn back(x: float)    -> int   { return x as int;   }  →  fptosi double %x to i64
fn widen(x: i32)     -> i64   { return x as i64;   }  →  sext i32 %x to i64
fn narrow(x: i64)    -> i32   { return x as i32;   }  →  trunc i64 %x to i32
fn fwiden(x: f32)    -> float { return x as float; }  →  fpext float %x to double
fn fnarrow(x: float) -> f32   { return x as f32;   }  →  fptrunc double %x to float
fn fromu8(x: i8)     -> i64   { return x as i64;   }  →  sext i8 %x to i64
fn fromi1(x: bool)   -> i64   { return x as i64;   }  →  zext i1 %x to i64
```

## Validation

- `make test-unit` — **86 / 86 passed**
- `compiler/tests/e2e/test_numeric_cast.sh build/native/sailfin` — **8 / 8 passed**
- `make test-integration` — **17 / 17 passed**
- Self-host stays green. Seed pin unchanged at `0.5.10-alpha.8`.

## Seed-DCE workaround (in-source)

The matrix opens with a no-op `coerce_operand_to_type(lowered.operand, lowered.operand.llvm_type, …)` keepalive call before the `let operand = lowered.operand` binding. Without it, the alpha.8 seed's dead-code pass elides the binding and every subsequent `operand.llvm_type` read returns garbage — the matrix never matches and all casts fall through to the generic "unsupported cast" diagnostic. `lower_return_instruction` in `statement.sfn` works around the same DCE quirk by passing `operand` to `coerce_operand_to_type` immediately after binding, which is the exact precedent followed here. Documented inline.

## L2/L3 silent-widening rejection: deferred to Slice E

The architect's plan paired the cast matrix with a `dominant_type` tightening that would refuse silent int↔float coercion. **Empirically that tightening cannot ship before Slice E retires `number`**: the lowered compiler/runtime code mixes i64 (from runtime helpers like `string.length`) and double (`number` alias) pervasively, so the rejection breaks user code like `for i in 0..arr.length` with no semantically meaningful fix-it. The LL evidence: with the tightening active, that loop's `i >= values.length` comparison harmonise-bailed, the if-check returned null, the break branch dropped, and the loop ran past the array length, producing `bounds_check failed: index=N length=N` at runtime. Slice D ships the `as` cast escape valve so authors can spell the conversion today; the implicit-rejection path rides on Slice E. Docs (`runtime_architecture.md` §3.7 + `status.md` row) updated to make this sequencing explicit.

## Test plan

- [x] `make compile` — self-host green
- [x] `make test-unit` — 86/86 passed
- [x] `make test-integration` — 17/17 passed
- [x] `compiler/tests/e2e/test_numeric_cast.sh build/native/sailfin` — 8/8 passed
- [x] All 8 cast pairs verified against direct `sfn emit llvm` smoke
- [ ] CI — pending push

## What changed since the reverted commit

- **Removed:** `dominant_type` int↔float sentinel, `harmonise_operands` diagnostic, comparison RHS-type propagation, `fs.exists` shim, `![io]` declaration, `.seed-version` bump.
- **Kept:** cast matrix, shared LLVM-type predicates, tests, docs (with deferral language).
- **Added:** `coerce_operand_to_type` keepalive (mirrors `lower_return_instruction`), updated docs/comments to be honest about Slice E sequencing.

https://claude.ai/code/session_01LKU4TzgTBtRM51D8x75UGN


mcereal:
Reverted in 344622c. CI surfaced widespread failures across `array_literal_aggregate_seed_test`, `arrays_test` (assertion failures), `build_cache_test`, `build_report_test`, `capsule_artifact_test`, `capsule_dep_resolution_test`, `check_tool_test` (`bounds_check failed: index=0 length=0`), `cli_path_normalization_test`, `cli_version_test`, `control_flow_test`, `cross_module_array_test`, and more. Mix of timeouts (124), assertion failures (134), and segfaults (139). Reproduced locally — `compiler/tests/unit/deep_recursion_test.sfn` hangs at `_repeat_string("ab", 1)`, which is a one-step return with no recursion involvement; the LLVM IR for the function looks correct in isolation, so the regression is somewhere else in the runtime/compiled binary that the dominant_type tightening reshaped.

**Root cause analysis.** My commit had two pieces:

1. **The cast lowering matrix in `lower_cast_expression`** — emits `sitofp`/`fptosi`/`sext`/`zext`/`trunc`/`fpext`/`fptrunc` per cast pair. Self-contained and verified working via the e2e shape pinning in `test_numeric_cast.sh`. **This piece is sound.**

2. **The `dominant_type` tightening in `core_operands.sfn`** — returns an empty sentinel when one side is integer (i1/i8/i16/i32/i64) and the other floating-point (double/float); `harmonise_operands` emits a diagnostic and bails with null operands. **This piece is what broke everything.**

The architect's plan said `dominant_type` was self-host-safe because "no in-tree code mixes annotated `int`/`float`." That's technically true at the **source** level. But the **lowered** code mixes integer-shaped and double-shaped operands constantly via the L1 path (bare-literal defaulting to `double`) and the integer-literal short-circuit at `core.sfn:460-491`. Concrete example: `arr.length > 0` lowers as `length: i64` (from the array struct field) compared against literal `0` (which lands as `double 0.0` because `expected_type` is empty in the comparison path). With the old `dominant_type`, that silently widens the i64 to double via `sitofp`. With my tightening, harmonise bails, the comparison returns null, and the binary lowering emits broken IR that compiles (because the runtime helper `bounds_check(i64, i64)` still gets called, but with garbage operands or zeroed defaults) and then misbehaves at runtime — exactly the `bounds_check failed: index=0 length=0` we see in `check_tool_test`.

The exact same shape explains the timeouts: a comparison that should terminate a recursion (`if n <= 0 { return 0; }`, `if n == 1 { return s; }`) lowers with a broken predicate, the `if` doesn't branch the way the source says, and the function loops.

**Path forward.** The cast lowering matrix should ship on its own — it's purely additive, hits new branches only when reached via `parse_cast_expression`, and the e2e test confirms it works. The `dominant_type` tightening cannot ship until **after** Slice E (bare-literal defaulting + `number` retirement) — without Slice E, the implicit int↔float mixing throughout lowered code is part of the language, not a footgun. Trying to close L2/L3 before then breaks the language.

The `fs.exists("/tmp/.sfn-seed-dce-keepalive/...")` shim I introduced for the seed-DCE workaround is also worth a hard look — the cast matrix worked locally with it, but the seed bug it papers over (the alpha.8/alpha.9 mis-indexing of `lowered.operand.llvm_type`) is real and should be tracked separately. Possible the shim itself is implicated in some of the runtime breakage, but the dominant_type tightening alone is sufficient to explain the failures we see.

Closing this PR draft. A clean re-attempt should:
1. Land the cast lowering matrix only (no `dominant_type` change).
2. Drop the seed-DCE shim if a cleaner pattern can be found, otherwise gate it behind the existing `debug_lowering` flag once we confirm the seed actually requires it (my earlier observation that the unconditional `print.err` pattern was needed may have been confirmation bias from a stale cache).
3. File a separate issue for the L2/L3 closure to ride on Slice E rather than Slice D.

Sorry for burning your CI minutes on this.

---

Issue #501, opened 2026-05-08
refactor(lowering): drop redundant seed_default_runtime_helpers carve-outs (M1.7.5a, closes #414)


mcereal:
## Goal

With #737 (the dynamic helper-collection precursor) merged AND a new seed cut that includes it, `collect_runtime_helper_targets` is the live source of `used_targets` flowing into `render_runtime_helper_declarations`. Prune the `seed_default_runtime_helpers` carve-out list in `compiler/src/llvm/lowering/lowering_helpers.sfn:681-` to only entries the dynamic walker provably cannot discover. **Closes #414.**

## Focus Workstream

Tracking: #321. Sub-issue of epic #494. Gated on #737 (precursor) being **in the pinned seed** — see "Required in pinned seed" below. The previous attempt at this issue (logged in https://github.com/SailfinIO/sailfin/issues/501#issuecomment-4527108530) failed precisely because the dispatch-collection mechanism the issue assumed was wired turned out to be hard-disabled in `lowering_core.sfn`. #737 fixes that.

## Scope

**In:**

- Edit `compiler/src/llvm/lowering/lowering_helpers.sfn:681-` (the body of `seed_default_runtime_helpers`).
- **Starting removal set** — the 14-entry candidate list from the prior empirical sweep (each previously failed the IR-diff gate *only because* the dynamic walker was dead; with #737 live, the dedup absorbs them naturally):
  - `copy_bytes`, `get_field`, `grapheme_at`, `mark_persistent`, `number.to_string`, `runtime.bounds_check`, `runtime_log_execution_fn`, `setjmp`, `sfn_exception_pop_frame`, `sfn_exception_push_frame`, `sfn_take_exception_frame`, `sfn_throw_frame`, `string.length`, `string.to_number`.
- Re-verify each candidate against an IR-diff sweep with #737 live. Entries that pass: delete. Entries that fail: re-add with an in-line comment stating *why* the dynamic walker can't discover it (e.g. *"module-level `let runtime_array_filter_fn = runtime.array_filter;` — not visible in any `NativeInstruction.expression`, see `lowering_helpers.sfn:719-723`"*).
- Document each **retained** carve-out in-line with the same WHY note. The seed-list already has good comments for the `sailfin_intrinsic_runtime_arena_*`, `nanosleep`, `fs.*` adapter, and `runtime_*_fn` module-binding cases — preserve and expand those.
- Reference #414 in the PR body so it auto-closes on merge.

**Out:**

[...]

PR #738, opened 2026-05-24
refactor(lowering): split skip_effects flag (helper-collection flip deferred to #740)

mcereal:
Refs #737 #740

## Status

Structural progress shipped. Helper-collection flip deferred to #740 (the determinism bug investigation).

## Summary

Three commits trace the diagnostic arc:

1. **`split skip_effects into helper-collection + effect-propagation flags`** (`lowering_core.sfn`, +10/-4) — the structural split exactly as scoped in #737. Replaces the single `skip_effects = true` flag with two independent flags: `skip_helper_collection` and `skip_effect_propagation`. Bootstrap-guard comment preserved verbatim on the effect-propagation side (citing commit `65b065ad`).

2. **`retire legacy *_exception walker carve-out (M2.7c)`** (`effects.sfn`, +13/-9) — and its revert. The M2.7c retirement was an attempt to neutralize the IR-diff drift on `try-catch-finally.sfn` that surfaced when the walker activated. It succeeded on that gate, but the walker-enabled state also surfaced a deeper determinism flake (#740) that blocked `make check`. With the walker deferred (commit 3 below), M2.7c is no longer needed; revert keeps the PR minimal.

3. **`defer helper-collection flip pending #740`** (`lowering_core.sfn`, +10/-3) — pins `skip_helper_collection = true`. Documents the determinism flake on `parser/expressions.sfn` (character drop + mangling-suffix loss in extracted call targets) and points to #740 for the fix.

Net effect: the structural split is in place so the eventual flip is a one-line change. No behavior change vs `main` until #740 lands.

## Acceptance Criteria (current state)

- [x] `grep "skip_effects = true" compiler/src/llvm/lowering/lowering_core.sfn` is empty (the variable was split into two).
- [x] **IR-diff invariant: byte-identical `emit-llvm` output** across all 17 fixtures (every `examples/basics/*.sfn` plus `compiler/tests/integration/async_await_test.sfn` and `compiler/tests/integration/closure_env_emission_test.sfn`). With the walker deferred, this is trivially satisfied — no behavior change vs `main`.
- [x] `sfn fmt --check` clean on both touched files.
- [x] `make compile` succeeds from seed `v0.7.0-alpha.3`.
- [x] `make test-unit` — 101/101 passed.
- [x] `make check` — full pipeline clean (linux-x86_64 locally verified; macOS CI parity expected after this push).

## Deviation from issue contract

The original issue's stated goal included "the runtime-helper auto-collection pass (`collect_runtime_helper_targets`) is restored to live." This PR delivers the structural prerequisite but **defers the actual flip** to #740 because re-enabling the walker reproducibly triggers a non-deterministic IR emission on `compiler/src/parser/expressions.sfn`:

```
< @expression_tokens_advance__parser__expressions
> @epression_tokens_advance
```

Both a character drop and the mangling-suffix loss arise only with the walker enabled. The 20x emit-sweep e2e test (`compiler/tests/e2e/test_runtime_array.sh`) catches this under `make check` load. The flake passes in isolation, fails reliably under the full check pipeline — symptom of a latent memory/aliasing bug exposed by the walker's additional allocations.

The deferral commit pins the flag and points the comment + commit message at #740 so the next picker has full context. The flip itself remains a one-line change.

## Files Changed

- `compiler/src/llvm/lowering/lowering_core.sfn` (+20, -7 net across 2 commits)
- `compiler/src/llvm/effects.sfn` (net zero — M2.7c added then reverted)

## Follow-up

- #740 — `bug(lowering): collect_runtime_helper_targets introduces non-deterministic IR on heavy modules`. Filed as `claude-ready` `size:m` `priority:high` `seed-blocker`. Repro and candidate root-cause surfaces are in the issue body.
- Once #740 closes: flip `skip_helper_collection = false`, re-apply (or re-derive) the M2.7c retirement (or its equivalent) to keep IR-diff zero, cut a fresh seed, `/pin-seed`, then #501 becomes pickable.

### DIFF
--- a/compiler/src/llvm/effects.sfn
+++ b/compiler/src/llvm/effects.sfn
@@ -547,21 +547,17 @@ fn collect_instruction_runtime_helper_targets(instruction: NativeInstruction) ->
         return helpers;
     }
     if instruction.variant == "Try" {
-        // M2.7c retirement (#737): legacy TLS-polling family
-        // (`clear_exception`, `has_exception`, `take_exception`) removed —
-        // they have no live emission sites in `compiler/src/` or
-        // `runtime/` (only the descriptor table in `runtime_helpers.sfn`
-        // still references them), so the M2.7b coexistence is a no-op in
-        // practice. Frame primitives are the only family the walker
-        // needs to declare for try-blocks.
-        return ["sfn_exception_push_frame", "sfn_exception_pop_frame", "sfn_take_exception_frame", "setjmp"];
+        // M2.7b (#404): emit declares for the new frame primitives alongside
+        // the legacy TLS-polling family so both APIs coexist during the
+        // transition. M2.7c retires the legacy entries once every emission
+        // site has cut over.
+        return ["clear_exception", "has_exception", "take_exception", "sfn_exception_push_frame", "sfn_exception_pop_frame", "sfn_take_exception_frame", "setjmp"];
     }
     if instruction.variant == "Throw" {
-        // M2.7c retirement (#737): legacy `set_exception` symbol removed
-        // alongside its Try-side counterparts above.
-        // `lower_throw_instruction` emits `call void @sfn_throw_frame`
-        // (longjmp); no live call sites reference the legacy name.
-        return ["sfn_throw_frame"];
+        // M2.7b (#404): `lower_throw_instruction` emits `call void
+        // @sfn_throw` (longjmp); declare the legacy `set_exception` symbol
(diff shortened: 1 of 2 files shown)


mcereal:
Update: `make test-unit` completed — **101/101 passed, 0 failed**. Functional gates are green; only the IR-diff invariant fails (set drift on `try-catch-finally.sfn` per the diagnosis in #737). `make check` was not run on the grounds that the IR-diff failure is dispositive for grooming, but happy to kick it if helpful.

---

Issue #832, opened 2026-05-27
feat(prelude): add Result<T, E> enum and Error struct


mcereal:
## Goal
Declare the canonical `Result<T, E>` generic enum and the default `Error` struct in the prelude, now that the pinned seed can lower a generic payload enum.

## Scope
**In:**
- `enum Result<T, E> { Ok { value: T }, Err { error: E } }`
- `struct Error { message: string; }`

**Out:**
- `?` type rules (R.3 — #833) and emitter (R.4 — #834)
- `E: Error` bound / `From<E>` coercion (deferred, needs generic constraints)

## Acceptance Criteria
- [ ] `enum Result<T, E> { Ok { value: T }, Err { error: E } }` added per Q3
- [ ] `struct Error { message: string; }` added per Q8
- [ ] make compile self-hosts (the NEW seed from #831 lowers the prelude `Result` correctly)
- [ ] A user `.sfn` can construct `Result.Ok { value: 5 }` and `match` it
- [ ] `prelude.ll` validates through the `llvm-as` cascade
- [ ] make test passes

## Files Affected
- runtime/prelude.sfn — add `enum Result<T, E>` and `struct Error`

## Verification
\`\`\`bash
ulimit -v 8388608 && make compile
ulimit -v 8388608 && timeout 60 build/native/sailfin test compiler/tests/unit/result_enum_lowering_test.sfn
\`\`\`

## Size
XS

## Type
Feature

## Blocked by
#831

## Required in pinned seed
#829, #830 (via the seed cut in #831)

## Context
- Epic #371 — R.2 split (R.2c). Isolated as XS because the hard work is in #829/#830; splitting makes the seed-gate explicit.
- CANNOT merge before #831 cuts and pins a seed containing #829+#830 — the seed compiles `runtime/prelude.sfn` and would otherwise type-erase the payload to `i8*`.
- `docs/runtime_architecture.md:944` already references `Result<SfnString, Error>` — this makes it real.

mcereal:
Unblocked: the sole blocker #831 is closed. The pinned seed (`0.7.0-alpha.16`) already lowers a generic payload-carrying enum correctly (verified via `result_enum_typecheck_test.sfn` 7/7 and `result_enum_lowering_test.sfn` 2/2 run directly on the seed), so `runtime/prelude.sfn` will no longer type-erase the `Result<T, E>` payload to `i8*`.

[...]

PR #899, opened 2026-05-29
feat(lowering): per-instantiation payload typing for generic payload enums

### LINKED ISSUE #830: feat(lowering): per-instantiation monomorphic layout for generic payload enums
## Goal
Emit a distinct LLVM struct layout and type name per concrete instantiation (`%"Foo<int,Error>"`) instead of type-erasing generic payload fields to `i8*`.

## Scope
**In:**
- Resolve generic payload field types from the instantiation context instead of the type-erasure fall-through
- Key enum layout on name + type-args, not name alone
- Per-instantiation payload GEP + field type in match lowering

**Out:**
- Prelude `Result` (R.2c — #831, gated on seed bump)
- `?` type rules (R.3) and emitter (R.4)

## Acceptance Criteria
- [ ] A standalone `.sfn` declaring `enum Foo<T,E> { Ok { value: T }, Err { error: E } }` and matching a `Foo<int,Error>` reads `value` as a by-value `i64`, not a loaded `i8*`
- [ ] `Foo<int,Error>` and `Foo<string,Error>` get distinct layouts/type names in the same module
- [ ] No regression on monomorphic enum tests (`enum_mixed_field_types_test.sfn`, `enum_match_struct_test.sfn`, `enum_complex_test.sfn`, `enum_state_machine_test.sfn`)
- [ ] New test `result_enum_lowering_test.sfn` compiles and runs (declares `Foo<T,E>` locally, NOT via prelude)
- [ ] make compile passes (self-hosts on current seed — the enum is in the test, not the compiler source)
- [ ] make test passes

## Files Affected
- compiler/src/emit_native_layout.sfn — `analyze_type_layout` :313, type-erasure fall-through :452 must resolve type params from instantiation
- compiler/src/llvm/type_context.sfn — enum-layout build :256-333 keyed on `enum_def.name` → must key on name+args
- compiler/src/llvm/lowering/lowering_phase_types.sfn — per-instantiation recovery
- compiler/src/llvm/lowering/instructions_match.sfn — payload GEP + `field_binding.field_type` :512
- compiler/tests/unit/result_enum_lowering_test.sfn (new)

## Verification
\`\`\`bash
ulimit -v 8388608 && make compile
ulimit -v 8388608 && timeout 60 build/native/sailfin test compiler/tests/unit/result_enum_lowering_test.sfn
\`\`\`

## Size
M

## Type
Feature

## Blocked by
#829

## Required in pinned seed
None

## Context
- Epic #371 — R.2 split. This is the issue whose code (with #829) must reach the pinned seed before #832 (prelude `Result`) can self-host.
- Enum layout is currently keyed on `enum_def.name` with a single `max_payload_size` — no `%"Result<int,Error>"` distinct from `%"Result<string,Error>"` exists anywhere.
- `examples/advanced/generic-structures.sfn` (`List<T>`) only "works" because every field is pointer-width; it never exercises by-value generic payloads.

mcereal:
## Summary
- Generic payload enums (`enum Foo<T, E> { Ok { value: T } }`) now resolve payload field types **per instantiation** at both `match` sites and the enum constructor. Previously a generic payload type-erased to the bare `%T`: match arms emitted `load %T` (rejected as an unsized load) and the constructor silently dropped the payload store, leaving the field uninitialized.
- Fixes a latent parser bug: `split_parameter_entries` ignored `<>` nesting, so a multi-arg generic parameter (`f: Foo<i64, i64>`) was comma-split into a truncated annotation **plus a phantom parameter**.
- New paths are gated on `type_parameters.length > 0` / a generic field, so **monomorphic enums are byte-for-byte unchanged**.

Closes #830

## Acceptance Criteria
- [x] A standalone `.sfn` declaring `enum Foo<T, E> { Ok { value: T }, Err { error: E } }` and matching a `Foo<i64,…>` reads `value` as a by-value `i64`, not a loaded `i8*` (verified in IR: `load i64` replaces `load %T`)
- [x] `Foo<int, int>` and `Foo<string, int>` coexist correctly in the same module (new test, second case)
- [x] No regression on monomorphic enum tests (`enum_mixed_field_types_test`, `enum_match_struct_test`, `enum_complex_test`, `enum_state_machine_test` — all pass)
- [x] New test `result_enum_lowering_test.sfn` compiles and runs (declares `Foo<…>` locally, **not** via prelude)
- [x] `make compile` self-hosts (the generic enum is in the test, not compiler source)
- [x] `make test` passes

## Design note (please read)
This implements per-instantiation **payload typing** over the backend's existing **shared opaque-payload `%Foo` layout**, rather than emitting distinct `%"Foo<…>"` named types. The match side derives the concrete payload type from the subject's declared source annotation; the construction side derives it from the lowered value operand (narrowing a fat `{i8*, i64}` string to its `i8*` data pointer). The two agree (`int`→`i64`, `string`→`i8*`).

This is correct for **pointer-width payloads** (`int`, `string`, boxed user types) — which covers prelude `Result<T, E>`'s needs. It does **not** yet emit distinct per-instantiation layout *sizes*: a generic field bound to a by-value type wider than 8 bytes would not get a larger payload slot. That refinement (true monomorphized layouts with mangled type names) is deferred; the de-erasure that unblocks the R.2 → #832 chain is delivered here.

## ⚠️ Discovered blocker for #832 (follow-up filed)
`sfn fmt` corrupts generic type annotations in **let-binding position**: `let r: Result<int, Error> = …` is reformatted to `Result<int, Error >= …` (the `> =` adjacency collapses into `>=`), producing semantically broken code. No existing compiler source uses `let x: Generic<Concrete> = …`, so it was never hit. This blocks the *natural* `Result<T, E>` annotation syntax #832 will need. The new test sidesteps it by constructing values inline as call arguments (fmt-safe, `fmt --check` clean). A separate follow-up issue tracks the fmt fix.

## Verification
```bash
ulimit -v 8388608 && make compile          # self-hosts ✓
ulimit -v 8388608 && build/native/sailfin test compiler/tests/unit/result_enum_lowering_test.sfn   # PASS (2/2)
# full suite: unit 111/111, integration 28/28, e2e 4/4, capsules 18/18, e2e-sh 81/81
# sfn fmt --check clean on all touched .sfn files
```

## Files Changed
- `native_ir.sfn`, `native_ir_parser_defs.sfn`, `native_ir_api.sfn` — `NativeEnum.type_parameters` + header capture
- `native_ir_utils_parse.sfn` — `<>`-aware `split_parameter_entries`
- `llvm/types.sfn`, `llvm/type_context.sfn`, `llvm/imports.sfn`, `llvm/lowering/lowering_phase_types.sfn` — `EnumTypeInfo.type_parameters`
- `llvm/type_mapping.sfn` — instantiation parse/substitute helpers
- `llvm/lowering/instructions_match.sfn` + `instructions_match_condition.sfn` — per-instantiation match field typing
- `llvm/expression_lowering/native/core_literals_lowering.sfn` — per-instantiation constructor payload store
- `compiler/tests/unit/result_enum_lowering_test.sfn` (new), `docs/status.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_017izHixy82BCdpmkx3upD7e)_

### DIFF
--- a/compiler/src/llvm/expression_lowering/native/core_literals_lowering.sfn
+++ b/compiler/src/llvm/expression_lowering/native/core_literals_lowering.sfn
@@ -16,7 +16,7 @@ import {
 import {
     array_struct_type_for_element,
     ends_with_pointer_suffix,
-    layout_annotation_represents_user_value,
+    enum_inst_annotation_references_param,
     map_type_annotation,
     strip_pointer_suffix
 } from "../../type_mapping";
@@ -1149,6 +1149,22 @@ fn lower_enum_literal(parse: EnumLiteralParse, bindings: ParameterBinding[], loc
         loop {
             if field_index >= variant_info.fields.length { break; }
             let expected = variant_info.fields[field_index];
+            // #830: a generic payload field (declared `value: T`) type-erases
+            // to `%T` in `expected.llvm_type`. The concrete instantiation is
+            // not recoverable here (the construction's target arrives already
+            // mapped to its boxed LLVM type), so for a field that references a
+            // declared type parameter we take the field's concrete LLVM type
+            // from the lowered value operand instead. This keeps the payload
+            // store in lock-step with the match-arm read, which derives the
+            // same concrete type from the subject's instantiation annotation.
+            // Monomorphic enums have no type parameters, so `field_is_generic`
+            // is false and `field_llvm_type` keeps the pre-mapped value.
+            let field_is_generic = enum_inst_annotation_references_param(expected.type_annotation, enum_info.type_parameters);
+            let mut field_llvm_type: string = expected.llvm_type;
+            // For a generic field, lower the value with no type hint so its
+            // natural LLVM type surfaces; otherwise hint the declared type.
+            let mut value_type_hint: string = field_llvm_type;
+            if field_is_generic { value_type_hint = ""; }
             let mut literal_index: int = -1;
             let mut literal_lookup_index: int = 0;
             loop {
@@ -1162,7 +1178,7 @@ fn lower_enum_literal(parse: EnumLiteralParse, bindings: ParameterBinding[], loc
 
             if literal_index >= 0 {
                 let literal_field = parse.fields[literal_index];
-                let lowered = lower_expression(literal_field.value, bindings, locals, current_temp, current_lines, functions, context, expected.llvm_type);
+                let lowered = lower_expression(literal_field.value, bindings, locals, current_temp, current_lines, functions, context, value_type_hint);
                 let mut _ci7: int = 0;
                 loop {
                     if _ci7 >= lowered.diagnostics.length { break; }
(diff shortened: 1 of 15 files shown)


mcereal:
Addressed in 5e15471: `make_native_enum` (`compiler/src/llvm/lowering/lowering_native_helpers.sfn:124`) now sets `type_parameters: []` explicitly. (In practice the omission didn't break the build — that constructor backs the dead `recover_native_enums_light` fallback the phase-4 arena path no longer arms — but the explicit default keeps every `NativeEnum` construction site consistent and prevents a silent #830 regression if the fallback is ever re-armed.) Same commit documents the `->`-inside-generic depth asymmetry in `split_parameter_entries` so the `depth > 0` guard isn't "fixed" into breaking the comma-split cancellation.

---

Issue #631, opened 2026-05-16
fix(driver): sfn build exits 0 when llvm lowering [fatal] fires on a void-return call


mcereal:
## Goal

Propagate `llvm lowering [fatal]` into `sfn build`'s exit code and abort artifact emission. Today the driver prints the fatal, then keeps going — for void-return call sites it actually writes an output binary and returns `0`, so a calling shell script (or CI) treats the build as successful.

## Current state

Discovered during #617 implementation (PR #630). When the lowering descriptor for a callee is missing, lowering emits:

```
llvm lowering [fatal]: cannot resolve return type for call to `spawn` — callee signature is not known to the compiler
```

For value-returning callees (e.g. `channel`, which previously returned `any`) the driver bubbles up rc=1 and skips the output binary, which is the intended behavior. For void-returning callees (e.g. `spawn`) the SAME diagnostic fires but the driver continues to the C-object link step, writes the output, and prints `built: <path>` with rc=0. Reproducer in `compiler/tests/e2e/test_sync_rejects_unimplemented_concurrency.sh` (the test was deliberately relaxed to *not* assert rc≠0 specifically because of this divergence — see its top-of-file comment).

This breaks the headline claim that the compiler fails closed on lowering errors. Any shell-script or CI gate that uses `sfn build`'s exit code as the contract gets a false green when a void-returning helper falls off the descriptor table.

## Scope

**In:**
- Trace `lowering_fatal` emission in `compiler/src/llvm/expression_lowering/native/core_call_resolution.sfn` (or wherever the "cannot resolve return type" string is generated — likely the same file that owns the void-vs-value branch) and the surrounding lowering driver, find the point where the void branch returns without propagating a failure flag back to `compile_to_llvm` in `compiler/src/main.sfn`.
- Plumb a single failure flag (or non-zero return code from the recursive lower fns) through to the entry point so any `[fatal]` aborts emission and returns non-zero from `sfn build` / `sfn emit llvm` / `sfn emit native`.
- The driver must not write `program` / the requested output path when a fatal fired.
- Promote the relaxed assertions in `compiler/tests/e2e/test_sync_rejects_unimplemented_concurrency.sh` back to the strict form: rc≠0 AND absence of the output binary. Remove the `|| true` and the "deliberately do NOT assert" comment block in `assert_build_rejects`.
[...]

Issue #616, opened 2026-05-16
investigate(typecheck): undefined function identifiers silently accepted


mcereal:
Sub-issue of #613.

## Goal

Determine whether the compiler should reject calls to undefined function identifiers, then either implement rejection or document the intentional auto-injection / `extern`-stub policy.

## Repro

```
$ cat > /tmp/undef.sfn << 'EOF'
fn main() {
    notarealfunction("hello");
}
EOF
$ ulimit -v 8388608 && build/native/sailfin check /tmp/undef.sfn
checked 1 files: ok      ← exit=0
```

Reproduced on Linux x86_64 with `sfn 0.5.10-alpha.19+dev`. Issue #549 mentions "auto-injected builtins (`print`, `assert`, basic types)" so some loose name resolution is intentional, but `notarealfunction` shouldn't be one of those.

## Scope

**In:**
- Audit `typecheck.sfn` Call handling — does it look up the callee in any symbol table?
- Identify whether the silence comes from missing name resolution, an `extern` carve-out, or auto-injection scope creep.
- Decide policy: (a) reject undefined identifiers at typecheck, (b) reject at link/lowering, (c) document the intentional behavior.

**Out:**
- Method / member resolution on objects (separate path, not affected here).
- Imported-symbol resolution (Phase E shipped, works).

## Acceptance Criteria

- [ ] A design doc or comment thread on this issue settles policy.
- [ ] If we reject: `notarealfunction("hi")` produces a clear `error[EXXXX]: undefined function 'notarealfunction'` diagnostic at typecheck or lowering time.
- [ ] If we don't reject: a passing test asserts the auto-injection envelope and `examples/README.md` mentions the policy.
- [ ] Existing examples (51 files per #549) still build and run.

## Files Affected (estimate)

- `compiler/src/typecheck.sfn` — likely the gap site.
- `compiler/src/emit_native.sfn` / `compiler/src/llvm/lowering/*` — may be where undefined symbols become silent link-time stubs.
- New unit test under `compiler/tests/unit/`.

## Size

- [ ] **needs-grooming** — needs architectural decision before implementation can be sized.

## Type

- [x] **Bug** (pending policy decision; may become `type:docs` if rejection is out of scope).

## Notes

Why this matters: an effect-system language whose typechecker doesn't notice the difference between `print.info(...)` and `notarealfunction(...)` undermines the "compile-time enforcement" pitch even when the effect part works.

mcereal:
[...]

Issue #613, opened 2026-05-16
Epic: Compiler integrity gaps — close the headline effect-enforcement holes


mcereal:
## Goal

Close the gap between Sailfin's headline "compile-time capability enforcement" claim and what the shipped binary actually does. Audit (sfn 0.5.10-alpha.18, externally-driven) surfaced several issues that, taken together, mean the marketed effect-system gate is not fully load-bearing today.

## Why this matters

`llms.txt` and `CLAUDE.md` codify the project's own rule: "Don't ship unfinished safety claims — features are either enforced end-to-end or explicitly deferred. No 'parsed but unenforced' marketing." Each gap below violates that rule for a specific surface.

### Tier 1 — Headline effect-enforcement integrity
- **Bare `print(...)` call is not classified as `![io]`.** Member-form `print.info(...)` trips `E0400`; the Call+Identifier branch in `effect_checker.sfn` covers `spawn`/`sleep`/`serve` but not `print`/`console`.
- **`sfn check` / `sfn emit llvm` exit 0 on errors (macOS arm64).** Diagnostics print but exit code is 0; summary shows garbage integers (`checked 22884130224 files: 22884131104 error`). Not reproduced on Linux x86_64 with the same source — suspected Bool/int ABI hazard across module boundaries (cf. the `find_char` double-ABI fix in commit `027735d9`).
- **Undefined call targets are silently accepted.** `notarealfunction("hi")` exits 0 with no diagnostic. Worth verifying whether this is an intentional auto-injection / `extern` hole or a real name-resolution miss.

### Tier 2 — Parsed-but-unenforced surface
- **`sfn/sync` `channel` / `parallel` / `spawn` are callable today but the C runtime symbols return `NULL` / no-op.** Typed `spawn_number`/`_string`/etc. variants are real pthread implementations; only the untyped public surface is a stub.

### Tier 3 — Test-suite coverage holes
- **`result_types_test.sfn` assertions are `assert 1 == 1`** (blocked by #50: match on union return types segfaults).
- **`check_tool_test.sfn`, `selfhosting_patterns_test.sfn`, `effect_gate_test.sfn` test local re-implementations**, not the compiler's actual exported symbols. Comment in `check_tool_test.sfn:1-13` documents the test-runner linker limitation that forces this.

### Tier 4 — Architectural carry-over
[...]

PR #1002, opened 2026-06-02
fix(driver): sfn emit native fails closed on unresolved-callee programs

### LINKED ISSUE #906: fix(driver): sfn emit native exits 0 on unresolved-callee programs that build/emit-llvm reject
## Goal

Make `sfn emit native` fail closed on the same unresolved-callee inputs that `sfn build` and `sfn emit llvm` now reject (after #631). Today `emit native` writes a `.sfn-asm` artifact and exits `0` for programs whose calls cannot be resolved, so any tool gating on `emit native`'s exit code gets a false green.

## Current state

Discovered while implementing #631 (PR pending). #631 made the LLVM `[fatal]` gate load-bearing: `sfn build` and `sfn emit llvm` now exit non-zero and write no artifact when a call's return type cannot be resolved (e.g. `spawn(...)` after its lowering descriptor was removed, or `let _ = channel(0)`).

`sfn emit native`, however, stops at **stage 5** — the `.sfn-asm` native-IR emitter (`write_native_text_file_with_module` / `compile_to_native_text_with_module` in `cli_main.sfn`) — which runs *before* **stage 6** LLVM lowering, where the `cannot resolve return type for call to ...` `[fatal]` is produced. As a result:

```bash
# value-returning unknown callee
printf 'import { channel } from "sfn/sync";\nfn main() ![io] { let _ = channel(0); }\n' > /tmp/v.sfn
sailfin emit -o /tmp/v.asm native /tmp/v.sfn ; echo rc=$?   # rc=0, /tmp/v.asm written

# void-returning unknown callee
printf 'fn main() ![io] { spawn(0, "x"); }\n' > /tmp/x.sfn
sailfin emit -o /tmp/x.asm native /tmp/x.sfn ; echo rc=$?   # rc=0, /tmp/x.asm written
```

This is uniform across value- and void-returning callees — `emit native` simply does not participate in the return-type resolution that the `[fatal]` gate depends on. #631 left this criterion unmet by design and deferred it here rather than expand into native-IR-stage resolution.

## Scope

**In:**
- Decide and implement the right layer for `emit native` to fail closed on unresolved callees. Two candidate approaches — the groomer/architect should pick:
  1. Run (or reuse) the callee-signature resolution at the native-IR stage so `emit_native.sfn` itself surfaces an unresolved-callee diagnostic and `cmd_emit` returns non-zero.
  2. Promote the relevant resolution check earlier (typecheck) so all three commands inherit the rejection from a shared gate — note this overlaps with #616 (`sfn check` doesn't flag undefined identifiers); coordinate so the two don't diverge.
- Add an e2e fixture asserting `sfn emit native` exits non-zero and writes no `.sfn-asm` for both a value- and a void-returning unresolved callee.

**Out:**
- Re-litigating the #631 `build` / `emit llvm` behavior (already landed).
- The `sfn check` typecheck silent-accept gap (#616) beyond coordinating the shared gate, if approach 2 is chosen.

## Acceptance Criteria

- [ ] `sfn emit native` on a program with an unresolved value-returning call exits rc≠0 and writes no `.sfn-asm`.
- [ ] Same for an unresolved void-returning call.
- [ ] Behavior matches `sfn build` / `sfn emit llvm` on identical inputs.
- [ ] `make compile` self-hosts on the unchanged seed.
- [ ] `make test` passes (unit + integration + e2e + capsules).
- [ ] `sfn fmt --check` clean on every touched `.sfn` file.

## Files Affected

- `compiler/src/emit_native.sfn` and/or `compiler/src/cli_main.sfn` — depending on chosen layer.
- `compiler/src/typecheck.sfn` — only if approach 2 (shared earlier gate).
- `compiler/tests/e2e/` — new fixture.

## Verification

```bash
ulimit -v 8388608
make compile
printf 'fn main() ![io] { spawn(0, "x"); }\n' > /tmp/x.sfn
build/native/sailfin emit -o /tmp/x.asm native /tmp/x.sfn ; echo "rc=$?"
test ! -e /tmp/x.asm && echo OK
```

## Size

S

## Type

Bug

## Required in pinned seed

None — internal to the native-IR emitter / driver exit-code paths the seed already executes.

## Blocked by

None. (#631 lands the `build` / `emit llvm` half this builds on; this issue covers the remaining `emit native` surface.)

## Context

- Split out of #631 acceptance criterion 3 (the `emit native` half), per maintainer direction, to keep #631 scoped to the lowering-stage `[fatal]` gate and avoid expanding into native-IR-stage resolution mid-session.
- Adjacent: #616 (`sfn check` doesn't flag undefined identifiers — typecheck layer). If approach 2 is chosen, coordinate so the shared gate covers both.
- Part of epic #613 (compiler integrity gaps — headline exit-code contract for LLM/CI loops).

mcereal:
## Summary

- `sfn emit native` stopped at the native-IR stage (stage 5), **before** the LLVM lowering stage (stage 6) where the `cannot resolve return type for call to ...` `[fatal]` is produced — so it wrote a `.sfn-asm` and exited `0` for programs `sfn build` / `sfn emit llvm` reject (after #631). A false green for any tool gating on its exit code.
- Adds a CLI-level lowering **dry-run gate** (`native_text_passes_lowering_gate`) that reuses the exact `[fatal]` machinery the other two surfaces use, so `emit native` now fails closed (rc != 0, **no** `.sfn-asm`) in parity with them. Writes no `.ll`.
- The gate runs **only** on the user-facing `emit native`. The build's per-module subprocess emit passes a new internal `--no-resolve-gate` flag to opt out: that stage carries no cross-module import context, so a legitimate imported call is unresolvable in isolation and must not be gated there — the build stays fail-closed via its later LLVM lowering stage (#631), not the native stage.

Closes #906

## Approach

Chosen layer (groomed at the design gate): **Approach 1 — a CLI-only lowering dry-run**, not Approach 2 (typecheck) and not a gate inside the native-emit workers. The decisive constraint: the self-host build uses the `emit native` subprocess as an intermediate per-module stage 5, where a single module has no cross-module signature visibility. A gate in the workers or at typecheck would fire on every legitimate cross-module call and break self-hosting (and entangle #616's unsolved cross-module-resolution problem).

## Acceptance Criteria

- [x] `sfn emit native` on a program with an unresolved value-returning call exits rc≠0 and writes no `.sfn-asm`.
- [x] Same for an unresolved void-returning call.
- [x] Behavior matches `sfn build` / `sfn emit llvm` on identical inputs.
- [x] `make compile` self-hosts on the unchanged seed.
- [x] `make test` passes (unit + integration + e2e + capsules).
- [x] `sfn fmt --check` clean on every touched `.sfn` file.

## Verification

```bash
# value-returning unresolved call
printf 'import { channel } from "sfn/sync";\nfn main() [io] { let _ = channel(0); }\n' > /tmp/v.sfn
build/native/sailfin emit -o /tmp/v.asm native /tmp/v.sfn ; echo rc=$?   # rc=1
test ! -e /tmp/v.asm && echo "no artifact OK"                            # no artifact OK

# void-returning unresolved call
printf 'fn main() [io] { spawn(0, "x"); }\n' > /tmp/x.sfn
build/native/sailfin emit -o /tmp/x.asm native /tmp/x.sfn ; echo rc=$?   # rc=1
test ! -e /tmp/x.asm && echo "no artifact OK"                            # no artifact OK

# parity / no-regression: known-good program still emits
build/native/sailfin emit -o /tmp/ok.asm native examples/basics/hello-world.sfn ; echo rc=$?  # rc=0, artifact present
```

Results: both unresolved cases → `rc=1`, no artifact; known-good → `rc=0`, artifact present.

- `make compile` — self-hosts (exit 0)
- `make test` — **e2e-sh 92/92, 0 failed** (exit 0)
- `sfn fmt --check` on all four `.sfn` files — clean
- `compiler/tests/e2e/test_sync_rejects_unimplemented_concurrency.sh` — 9/9 (incl. 2 new #906 cases)
- `compiler/tests/e2e/test_effect_gate_build_path_entry.sh` — 3/3 (guards the build path is not regressed by the gate)

## Files Changed

- `compiler/src/llvm/lowering/lowering_core.sfn` — `native_text_passes_lowering_gate` (side-effect-free dry-run).
- `compiler/src/main.sfn` — gated wrappers `write_native_text_file_with_module_gated` / `print_native_text_with_module_gated` (with a `run_gate` opt-out); bare `write_native_text_file_with_module` untouched (the build calls it directly).
- `compiler/src/cli_main.sfn` — `cmd_emit` native branch rewired to the gated wrappers + `--no-resolve-gate` flag.
- `compiler/src/emit_helpers.sfn` — build subprocess emit passes `--no-resolve-gate`.
- `compiler/tests/e2e/test_sync_rejects_unimplemented_concurrency.sh` — `assert_emit_native_rejects` + value (`channel`) and void (`spawn`) cases.

Part of epic #613 (compiler integrity gaps — headline exit-code contract for LLM/CI loops).

---
_Generated by [Claude Code](https://claude.ai/code/session_01VM7k1uEkcD4XcdQ3psJvFZ)_

### DIFF
--- a/compiler/src/cli_main.sfn
+++ b/compiler/src/cli_main.sfn
@@ -69,12 +69,12 @@ import {
     compile_to_llvm_file_with_module,
     compile_to_llvm_with_module,
     compile_to_llvm_with_module_timed,
-    compile_to_native_text_with_module,
     compile_to_sailfin,
     module_name_from_path,
     number_to_string,
     print_llvm_with_module,
-    write_native_text_file_with_module
+    print_native_text_with_module_gated,
+    write_native_text_file_with_module_gated
 } from "./main";
 import { RuntimeCapsuleArtifacts, runtime_capsule_from_root } from "./runtime_capsule_resolver";
 import {
@@ -1499,6 +1499,11 @@ fn sailfin_cli_main(argv: string[]) -> int ![io, clock] {
         // round-trip.
         let mut emit_attempts: int = 1;
         let mut emit_validate: boolean = false;
+        // #906: build-internal per-module native emits set this (via
+        // `--no-resolve-gate`) to skip the lowering dry-run gate, which
+        // would otherwise reject legitimate cross-module calls that the
+        // import-context-free native stage cannot resolve in isolation.
+        let mut skip_resolve_gate: boolean = false;
         let mut base_index = 1;
         // Parse optional flags: --timing, --module-name <slug>,
         // -o <path>, --attempts N, --no-retry, --validate,
@@ -1577,6 +1582,12 @@ fn sailfin_cli_main(argv: string[]) -> int ![io, clock] {
                 consumed_flag = true;
                 continue;
             }
+            if strings_equal(args[base_index], "--no-resolve-gate") {
+                skip_resolve_gate = true;
+                base_index += 1;
+                consumed_flag = true;
+                continue;
+            }
         }
         let _expected_args = base_index + 2;
         if args.length != _expected_args {
@@ -1635,11 +1646,24 @@ fn sailfin_cli_main(argv: string[]) -> int ![io, clock] {
             return 0;
         }
         if strings_equal(mode, "native") {
+            // #906: gate the native emit on the same lowering `[fatal]`
+            // that `build` / `emit llvm` reject, so `emit native` fails
+            // closed (rc != 0, no `.sfn-asm`) on unresolved-callee inputs
+            // instead of writing an artifact and exiting 0. The build's
+            // per-module subprocess emit passes `--no-resolve-gate`
+            // (`skip_resolve_gate`) to opt out: that stage emits
+            // intermediate `.sfn-asm` with no cross-module import context,
+            // so it must not gate here — the build stays fail-closed via
+            // its later LLVM lowering stage (#631).
+            let run_gate = !skip_resolve_gate;
             if has_out {
-                let ok = write_native_text_file_with_module(source, module_name, out_path);
-                if !ok { return 1; }
+                if !write_native_text_file_with_module_gated(source, module_name, out_path, run_gate) {
+                    return 1;
+                }
             } else {
-                print(compile_to_native_text_with_module(source, module_name));
+                if !print_native_text_with_module_gated(source, module_name, run_gate) {
+                    return 1;
+                }
             }
             return 0;
         }
--- a/compiler/src/emit_helpers.sfn
+++ b/compiler/src/emit_helpers.sfn
@@ -247,7 +247,14 @@ fn emit_subprocess_with_retry(sailfin_exe: string, module_name: string, source_p
         // is always the same-generation binary (`<binary_dir>/sailfin`),
         // so it understands these flags; the seed never executes this
         // helper (pass 1 of `make compile` runs the seed's own resolver).
-        let mut emit_args: string[] = [sailfin_exe, "emit", "--module-name", module_name, "--attempts", "1", "--no-validate", "-o", out_path];
+        // `--no-resolve-gate` (#906): this is the build's per-module
+        // intermediate emit. For the `native` stage it carries no
+        // cross-module import context, so the lowering dry-run gate would
+        // falsely reject legitimate imported calls; suppress it here and
+        // let the build's LLVM lowering stage (#631) own fail-closed. The
+        // `llvm` stage ignores the flag. The same-generation `sailfin_exe`
+        // understands it; the seed never executes this helper.
+        let mut emit_args: string[] = [sailfin_exe, "emit", "--module-name", module_name, "--attempts", "1", "--no-validate", "--no-resolve-gate", "-o", out_path];
         if import_context_root.length > 0 {
             emit_args.push("--import-context");
             emit_args.push(import_context_root);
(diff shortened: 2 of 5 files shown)


mcereal:
CI note: `Build + Test [linux-x86_64]` failed on the `20x emit-sweep on heavy module is byte-identical` assertion in `test_runtime_array.sh` (iteration 8 diverged in a `%TokenKind` enum-variant `select` chain). This is the **known, tracked non-determinism flake #892** ("residual non-deterministic heap corruption in runtime-helper declare emission") — a latent string-buffer aliasing/UAF in the helper-collection→declare path that re-surfaces whenever heap layout shifts. `test_runtime_array.sh`'s emit-sweep is the named repro for it.

It is **unrelated to this PR's exit-code gating**: that test exercises `sfn emit llvm` (the `compile_native_text_to_llvm_file` path), whereas this PR only adds `native_text_passes_lowering_gate` (called solely by `emit native`) and an empty-text guard in the native stdout wrapper — neither alters lowering output or ordering. The PR's only connection is that an extra allocation shifts heap layout enough to occasionally trip #892. The `emit native` (#906) assertions themselves passed (9/9), and the sweep passes locally (4× consecutive here). Re-running the failed leg.

---

Issue #843, opened 2026-05-27
Epic: test-infra Phase 4 — Parallel test execution (CI sharding + per-test cache now; in-process --jobs runtime-gated)


mcereal:
**Parent epic:** #839 · **Phase:** 4 of 4 · **Status:** **active (pre-1.0)** — fast CI feedback is a prerequisite for reaching 1.0, not a post-1.0 nicety. A slow PR loop (Linux ~24 min, macOS ~47 min, ~90% of it the serial test runner) directly throttles iteration speed toward 1.0, so the wall-time goal is pulled forward.

## Reframing (2026-06-03)

This epic was previously deferred to post-1.0 because it was scoped as **in-process `--jobs N`** parallelism, which needs the `routine`/`spawn`/work-queue runtime that does not exist yet. That conflated two separable things:

1. **The wall-time goal** ("≥4× faster test execution") — achievable **pre-1.0 with no concurrency runtime** via CI-level matrix sharding (GitHub supplies the parallelism) + a content-addressed per-test binary cache. **Pickable now.**
2. **In-process `--jobs N`** — still genuinely gated on `process.spawn`/`routine`. Stays runtime-gated, but is no longer what blocks the *goal* or the epic.

Plan of record for the pre-1.0 path: [`docs/proposals/ci-test-speed.md`](https://github.com/SailfinIO/sailfin/blob/main/docs/proposals/ci-test-speed.md). Phase source of truth: [`02-phases.md` Phase 4](https://github.com/SailfinIO/sailfin/blob/main/docs/proposals/test-infra-epic/02-phases.md#phase-4).

### Track A — pre-1.0, no concurrency runtime (do now)

- [ ] **4.A1** CI matrix test sharding — fan unit/integration/e2e/capsules/e2e-sh across N concurrent jobs per OS via the runner's existing multi-path/file-list args (`cli_commands.sfn:919-937`) + a thin `make test-shard` seam + a **shard-cover lint** (union of shard file lists == full `find`). Pure CI-config; touches no `.sfn`. Target: macOS ~47→~16-18 min, Linux ~24→~10-11 min.
[...]

Issue #513, opened 2026-05-08
Epic: Makefile slim-down — sfn owns build / check / cross / stamp


mcereal:
## Goal

Collapse the (~1,155-line) root `Makefile` to a thin convenience wrapper (~80 lines, no heavy lifting) by promoting build / check / cross-compile / stamp / **runtime-link** responsibilities into first-class `sfn` subcommands. End state: every non-bootstrap recipe is at most three lines invoking `sfn`.

---

## ⚑ Status update — 2026-06-22 (re-groom)

Roughly **half the epic has shipped in the compiler** (verified against source, not just issue-closed). Two strategy shifts since the original grooming:

- **Phase 2 (cross-compile) is superseded.** The plan to fold the mingw cross-from-Linux path *into* `sfn build --target=` has been **reversed** by the new **Epic #1485 (Native Windows self-host, MSVC ABI)**: Windows binaries now come from a native MSVC self-host on a Windows runner, and **#1496 deletes `ci-cross-windows` + the `RUNTIME_MODS` loop outright** (~250 Makefile lines *removed*, not migrated). Phase 2 is therefore **dropped from this epic** and tracked under #1485.
- **Makefile end-state = thin wrapper (confirmed), not full deletion.** Bootstrap primitives (`fetch-seed`/`install`/`clean*`/`mcp-server`) stay as Make targets / `install.sh`.
- **`make fmt` dropped.** `sfn fmt` already exists; adding a Make target cuts against shrinking the Makefile. Any fmt-ergonomics work belongs in `sfn fmt`'s own defaults.

### Verified phase status

| Phase | Status | Evidence |
|---|---|---|
| **0.1** cache invalidation (#514) | ✅ **DONE** | `*.sfn-asm` wipes gone from Makefile. |
| **0.2** emit retry parity (#515) | ✅ **DONE** | `emit_llvm_with_retry` wired at `cli_main.sfn:2151`; `--validate`/`--attempts` flags. |
| **0.3** build-stamp (#516) | ✅ **DONE** | `build_stamp.sfn:107` called from `cli_main.sfn:2559/2623`; Makefile has zero stamp logic. |
| **R** runtime-link unification (#939) | ✅ **DONE** | `_clang_compile_runtime_objects` deleted; `_clang_link_multi_with_opt` (`cli_main.sfn:1659`) has no legacy `.o`-probe branch; all `_runtime_<mod>_path` helpers gone; runtime resolves as an implicit capsule dep. |
[...]

Issue #339, opened 2026-05-06
Tracking: Build Performance & Production-Readiness — 2026-05-06


mcereal:
## Goal

Lock the workstreams that take Sailfin's build pipeline from its current state (cold full self-build 2m27s on 4 cores; per-module peak RSS 1.3–1.8 GB on three modules; `EMIT_RETRIES=3` still load-bearing) to a production-ready, Rust/Go/Zig-class build experience before 1.0.

This is a tracking/coordination issue — the architectural decision and rationale. Implementation work fans out into the linked sub-issues.

## Context

A build-performance stocktake on 2026-05-06 (memory: `sailfin-design-decisions/build-perf-stocktake-2026-05-06.md`) confirmed the IPC-as-GC era is over (Phase 2 closed in 0.5.9, all flag-file probes retired #308/#311/#312), parallelism is paying out (Stage E PR3 cut cold-build 6m07s → 2m27s on 4 cores), and the cache layer is shipped (Stage C PR1f content-addressed + `--json` build report). Three structural gaps remain between today's pipeline and the 1.0 bar:

1. **Per-module RAM is 3–4× target** on three modules (`parser/declarations.sfn` 1759 MB, `lowering_recovery.sfn` 1570 MB, `instructions_match.sfn` 1361 MB). The `lowering_core.sfn` split (commits `072bea1`, `cd34d14`) demonstrated the playbook works (2350 → 1594 MB, –32%). Three more modules need the same treatment so `BUILD_JOBS=8` works on commodity 8 GB CI runners and macOS comes off the `BUILD_JOBS=2` cap.
2. **No CI gates on the cache or determinism.** Stage C PR1f shipped `BuildReport.cache.{hits, misses, stores}` but no PR step asserts hit rate on no-source-change reruns. `--check-determinism` is in the build-architecture proposal but not wired to PR CI. Both are silent regression risks.
3. **The compiler is monolithic.** Stage G (`sfn/compiler-parser`, `sfn/compiler-typecheck`, `sfn/compiler-emit`, `sfn/compiler-llvm`) was demoted to post-1.0 in `docs/proposals/build-architecture.md` §4.10 / §Stage G. **Per user direction (2026-05-06): this is a 1.0 deliverable, not post-1.0.** Doesn't have to land immediately, but the 1.0 release will not ship with the monolithic shape.

## What's affirmed (KEEP)

[...]

PR #1012, opened 2026-06-03
perf(ci): shard PR test suite across parallel legs (#843 Track A)

mcereal:
## Summary

PR CI is slow because the **test suite is ~90% of wall time and runs serially** — Linux ~22m / macOS ~42m of the ~24m/~47m job totals (measured from run 26894545759 step timestamps; the warm-cache build is only 1.5–3m).

This fans the suite across `os × shard` legs that run **concurrently**. Each leg builds the compiler (warm cache) then runs only its shard. No compiler/`.sfn` changes — self-host, determinism, and fmt gates are untouched.

- **`scripts/test_shards.sh`** — single source of truth for the shard→file map (`unit-a` / `unit-b` / `int-e2e-caps` / `e2e-sh`). `cover` asserts the union equals `make test`'s surface (175 `*_test.sfn` + 94 `test_*.sh`, balanced 59/59/57, no gaps or overlaps).
- **Makefile** — `make test-shard SHARD=<name>` and `make test-shard-cover`.
- **`ci.yml`** — matrix 2 → 2×4 legs; a cheap **`shard-cover` gate job** fails fast before the expensive build legs; one-shot steps (fmt / canary / effect-gate / cross-compile / package / upload) run only on the **primary leg per OS** (`unit-a`) so they don't run 4×.
- **`sailfin-build` action** — new `shard` / `package` inputs **default to current behavior**, so `release-tag.yml` and `release-branches.yml` are unaffected.

Expected: **macOS ~47 → ~16–18m, Linux ~24 → ~10–11m, zero coverage lost.** This PR runs under the new sharded CI, so its own run is the measurement.

Costs `(N−1)×` redundant build runner-minutes (Phase 1a, accepted) — Phase 2 shares the build artifact across shards. Per-test binary cache (Lever 2) and scope-trim (Lever 3) follow per the plan.

Re-prioritizes **#843 to pre-1.0**: Track A (this PR — CI sharding) achieves Phase 4's wall-time goal with **no `routine`/`spawn` runtime**; in-process `--jobs` (Track B) stays runtime-gated.

Plan of record: `docs/proposals/ci-test-speed.md`.

Refs #843, #513, #339

## Test plan

- [ ] `shard-cover` job is green (union == `make test` surface)
- [ ] All 8 build legs green; `unit-a` legs run fmt/canary/effect-gate (+ Linux cross-compile/uploads) exactly once per OS
- [ ] Aggregate pass count across shards == a baseline `make test` pass count on the same SHA
- [ ] macOS critical-path wall time drops to ~16–18m; Linux to ~10–11m
- [ ] Release workflows unaffected (composite defaults: full `make test` + packaging)

### DIFF
--- a/.github/actions/sailfin-build/action.yml
+++ b/.github/actions/sailfin-build/action.yml
@@ -130,6 +130,15 @@ runs:
       shell: bash {0}
       run: |
         set -euo pipefail
+        # Memory cap (.claude/rules/compiler-safety.md): on Linux, cap each
+        # compiler invocation at 8 GB virtual so a runaway compile (e.g. a
+        # #892 corruption hitting a bad allocation) fails *cleanly* per-test
+        # instead of OOM-killing the whole hosted runner ("runner received
+        # shutdown signal"). ulimit is inherited by `make` and every test
+        # child it forks. Darwin does not honor RLIMIT_AS, so skip on macOS.
+        if [ "${RUNNER_OS:-}" = "Linux" ]; then
+          ulimit -v 8388608 || true
+        fi
         # Empty shard = full suite (release workflows + back-compat).
         # A shard name runs only that disjoint slice; the union across
         # legs equals `make test` (asserted by the shard-cover lint).
--- a/.github/actions/sailfin-build/action.yml
+++ b/.github/actions/sailfin-build/action.yml
@@ -133,10 +133,24 @@ runs:
         # Empty shard = full suite (release workflows + back-compat).
         # A shard name runs only that disjoint slice; the union across
         # legs equals `make test` (asserted by the shard-cover lint).
-        if [ -n "${{ inputs.shard }}" ]; then
-          make test-shard SHARD="${{ inputs.shard }}"
-        else
-          make test
+        run_tests() {
+          if [ -n "${{ inputs.shard }}" ]; then
+            make test-shard SHARD="${{ inputs.shard }}"
+          else
+            make test
+          fi
+        }
+        # Retry-once: #892 is a live non-deterministic codegen heap bug
+        # that intermittently corrupts an *unrelated* test's compile
+        # (garbled link temp name, or a hang -> timeout). A re-run hits a
+        # different random victim and almost always passes. A *persistent*
+        # failure still fails the retry -> red, so real breakage is not
+        # masked. The ::warning:: keeps every flake hit visible/tracked
+        # in the run summary so #892's frequency stays measurable.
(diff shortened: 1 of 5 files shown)


mcereal:
## Review follow-up + failure diagnosis

### Copilot inline comments — addressed in eed43992
- **Makefile:286 / :291** (invoke via `bash`): done. `test-shard` and `test-shard-cover` now call `bash scripts/test_shards.sh`, matching `detect_build_jobs.sh` / `bench_compile.sh` and removing the exec-bit dependency. (FWIW the file is committed `100755`, so it wasn't actually broken — but the consistency point is fair.)
- **ci-test-speed.md:4** (stale status): done — the status line now says Phase 1 is implemented.

### Copilot's low-confidence matrix concern — empirically refuted
The suppressed comment worried that `include` entries specifying only `os` wouldn't merge `package_label`/`clang` into the `os × shard` combinations (and would instead spawn rows with `shard` unset). The live run shows otherwise — the matrix expanded to **exactly 8 legs**, each correctly named and typed:

```
Build + Test [macos-arm64 / unit-a]      Build + Test [linux-x86_64 / unit-a]
Build + Test [macos-arm64 / unit-b]      Build + Test [linux-x86_64 / unit-b]
Build + Test [macos-arm64 / int-e2e-caps] ...
```

GitHub merges an `include` entry whose matrix-axis keys (`os`) match into the existing combinations (adding `package_label`/`clang`) rather than creating new rows, because `shard` is still set on those combinations. No phantom rows; `target` was populated on every leg.

### The `linux-x86_64 / unit-b` failure is pre-existing #892, not sharding
```
/usr/bin/ld:/tmp/sfn-test-.../sub-31<garbled>: file format not recognized; treating as linker script
clang: error: linker command failed with exit code 1
```
A non-ASCII byte spliced into an intermediate object-file name is the string-buffer corruption signature of **#892** (residual non-deterministic heap corruption in runtime-helper `declare` emission). Two facts confirm it's the flake, not the shard map:
1. **`macos-arm64 / unit-b` passed** — the *same shard file set* is green on the other OS.
2. Sharding only selects which files a leg runs; it can't influence codegen temp-file naming.

`linux-x86_64 / int-e2e-caps` was *cancelled* mid-build (not failed) with `fail-fast: false` and no superseding run — a transient runner reclaim.

The push of eed43992 re-runs all legs. Expecting green absent another #892 hit. Tracking the reliability angle separately — sharding is subject to #892's intermittency, so #892 matters for CI reliability.

PR #1014, opened 2026-06-03
chore(seed): pin seed to 0.7.0-alpha.22

mcereal:
## Summary
Bump the self-hosting seed pin from `0.7.0-alpha.21` → `0.7.0-alpha.22` (the latest released compiler, matching `main` commit 3b36786f).

## Verification
- `make fetch-seed` downloaded `sailfin_0.7.0-alpha.22_linux_x86_64.tar.gz` successfully
- `build/seed/bin/sailfin version` → `sfn 0.7.0-alpha.22` ✅

## Notes
- Only `.seed-version` changes; no build artifacts committed.
- Full `make compile` from the new seed (~13 min) was **not** run locally — recommend confirming green CI (which builds the compiler from this pin) before merge.


mcereal:
Closing: the `0.7.0-alpha.22` seed cannot self-host current `main`. Every Build+Test leg (both platforms) fails in the parallel `stage_import_context` pass with `llvm lowering [fatal]: cannot resolve return type for call to ... — callee signature is not known to the compiler` across core cross-module callees (`parse_program`, `find_char_local`, `lift_non_capturing_lambdas`, the `_*_cmd` helpers), plus int/float ABI-mismatch fatals. Reproduced locally via `make rebuild`.

Likely root cause: interaction of #1001 (cross-module callee signature resolution on `-p` binary entry) and #1002 (emit fails *closed* on unresolved callees) — alpha.21 fails *open* and silently emits, which is why main is green today. Tracking the regression separately; will re-pin once a good seed exists.

Issue #490, opened 2026-05-08
refactor(numeric): migrate prelude + tests, reapply #296 (Slice E.3)


mcereal:
## Goal

Close Slice E by migrating `runtime/prelude.sfn` and `compiler/tests/` off bare `number` for integer-shaped values, then reapply #296 (`dominant_type` rejection of silent int/float coercion) on top of the now-aligned tree. After this slice the prelude, tests, and compiler source all agree on `int` vs. `float`; bare `number` survives only in three opt-out spots (the `Future<number>` mapping seam, literal-text predicates in `core_text.sfn`, and the language reference where it is documented as a deprecated alias for `float`).

## Focus Workstream

Tracking #321, M0 prerequisite under epic #424 (Slice E). E.3 is the closing slice; depends on E.1 (language-level rule + alias) and E.2 (compiler source migration).

## Scope

**In:**
- Migrate `runtime/prelude.sfn` (7 hits) off `: number` for integer-shaped parameters and returns. Specifically: `sleep(milliseconds)`, `monotonic_millis()`, `channel<T>(capacity)`, `char_at(index)`, `descriptor_find_top_level`, `clamp(value, min, max)` (decide int vs. float per call site; current usage is integer indexing and counts), `substring(start, end)`, `find_char(start)`, `char_code` return, `grapheme_count` return, `grapheme_at(index)`. Verify each by reading the call sites in `compiler/src/`.
- Migrate `compiler/tests/unit/`, `compiler/tests/integration/`, `compiler/tests/e2e/` (~97 files containing `: number`). Tests asserting `: number` in parser/typecheck shape should split: tests pinning the `int` default land alongside; tests pinning legacy `number`-as-float behaviour (e.g. `let x: number = 1.5`) stay but get a comment marking them as alias-coverage.
- Reapply #296: `dominant_type` rejects silent int/float coercion at `compiler/src/llvm/expression_lowering/native/core_operands.sfn:1642`. With both sides of every site agreeing post-E.2/E.3, the rejection no longer breaks `for i in 0..arr.length`-style code. Restore the diagnostic plus fix-it suggestion ("add `as float` or `as int` to disambiguate") that was reverted on branch `claude/slice-d-cast-lowering-cWzX2` commit `ffd2620`.
[...]

PR #1131, opened 2026-06-06
numeric: classify i1 as a distinct "bool" kind, refuse silent bool→numeric widening

### LINKED ISSUE #537: tech-debt(numeric): tighten boolean (i1) numeric_type_kind classification post-Slice E
## Goal

Revisit the `numeric_type_kind` classification to either treat boolean (`i1`) as `"int"` for ABI-shape diagnostic purposes, or introduce a separate `"bool"` kind with explicit coercion rules. After this issue lands, silent `i1 → double` (and `i1 → i64`) widening at interpolation, print, and arithmetic boundaries either becomes explicit (`as int` / `as bool`) or is documented as deliberately tolerated.

## Focus Workstream

Tracking #321, post-M0 Slice E follow-up. Sub-issue under epic #424 (Slice E) but **deliberately deferred** to after #490 (E.3) closes. This is `needs-grooming` — not picked up until the broader Slice E lands and the architect re-evaluates the priority.

## Background

#528 (Slice E.1.5) introduced `numeric_type_kind(llvm_type) -> "int" | "float" | "other"` as the chokepoint for ABI-shape diagnostics. During implementation the agent locked in **`i1` → `"other"`** (rather than `"int"`) because every interpolation/print path in the existing tree silently widens `i1 → double`. Treating bools as `"int"` would have cascade-fired diagnostics across the entire compiler tree.

This is a deliberate scope choice, locked in by the integration test in `compiler/tests/integration/numeric_abi_mismatch_test.sfn`. It works for E.1.5/E.2/E.3's purposes — booleans are not part of the integer-shape migration. But it leaves a long-term hole: the language has no compile-time check that `b: bool` flowing into a numeric context is intentional.

## Scope

**In:**
- Audit every site in `compiler/src/` and `runtime/prelude.sfn` that performs implicit `i1 → i64` or `i1 → double` widening. The dominant cases are interpolation (`print.info("count: ${b}")` where `b: bool`) and arithmetic (`if b { x + 1 } else { x }` patterns).
- Decide between two design options (architect call):
  1. **Treat `i1` as `"int"`** in `numeric_type_kind`, audit-and-fix every implicit widening site to use explicit `as int`. Stricter, larger diff.
  2. **Introduce a separate `"bool"` kind**, with diagnostic on `bool ↔ int` and `bool ↔ float` mismatches. Add explicit `as int` / `as bool` casts where needed. Cleanest semantically, requires teaching the language about `bool` as a first-class kind.
- Implement the chosen design with the appropriate fix-it diagnostics.
- Update the spec under `site/src/content/docs/docs/reference/spec/06-types.md` to document the new boolean coercion contract.

**Out:**
- Removing the `bool` keyword or changing its parsing — purely a coercion-tightening issue.
- Revisiting `numeric_type_kind`'s treatment of pointer or aggregate kinds (still `"other"`).
- Anything in the rest of Slice E (E.1.5 / E.2 / E.3).

## Acceptance Criteria

- [ ] Architect decision recorded on this issue (option 1 vs. option 2 vs. variant).
- [ ] Either `numeric_type_kind("i1") == "int"` (option 1) or a new `"bool"` kind exists (option 2), with diagnostics covering the chosen behaviour.
- [ ] Negative test: a synthesised `let b: bool = true; let n: int = b` compile fails with the new diagnostic (option 1: kind mismatch; option 2: explicit-cast suggestion).
- [ ] Every existing implicit `i1 → numeric` widening in `compiler/src/` and `runtime/prelude.sfn` has either been migrated to an explicit cast or annotated as deliberately tolerated.
- [ ] `make compile` self-hosts on the post-Slice-E seed.
- [ ] `make test` passes.
- [ ] `make check` green.
- [ ] `sfn fmt --check` clean.
- [ ] Spec update in `site/src/content/docs/docs/reference/spec/06-types.md`.

## Files Affected

- `compiler/src/llvm/type_mapping.sfn` — `numeric_type_kind` extension.
- `compiler/src/llvm/expression_lowering/native/core_operands.sfn` — diagnostic extension (or option-2 separate `bool` path).
- `compiler/src/llvm/expression_lowering/native/core_strings.sfn` — interpolation site (likely the densest implicit-widening hotspot).
- `compiler/tests/integration/numeric_abi_mismatch_test.sfn` (extend) or new `numeric_bool_coercion_test.sfn`.
- Various `compiler/src/*.sfn` files where implicit widening currently exists — count to be determined during the audit phase.
- `runtime/prelude.sfn` — same audit.
- `site/src/content/docs/docs/reference/spec/06-types.md` — coercion-rules section.

## Verification

```bash
ulimit -v 8388608
make compile
make test
make check
sfn fmt --check compiler/src/ runtime/

# Audit
# (option 1) confirm numeric_type_kind tests cover bool → int
# (option 2) confirm bool kind diagnostic fires on the negative test
```

## Size

S (post-architect-decision; could grow to M if the audit finds many implicit-widening sites).

## Type

Tech-debt

## Blocked by

#490 (Slice E.3). The broader integer/float tightening must finish before this slice can be evaluated meaningfully — the implicit-widening audit's result depends on whether E.3's changes already eliminated some of the i1 sites.

## Required in pinned seed

None — the audit works against any seed that includes #528's `numeric_type_kind` helper.

## Context / Background

- Parent epic #424 (Slice E). Predecessor: E.3 (#490).
- Tracking #321, M0 follow-up.
- Surfaced during #528 (Slice E.1.5) implementation. The agent's note: *"i1 (boolean) classified as `other`, not `int` — every interpolation/print path widens i1 → double today; treating bools as int kind would cascade-fire diagnostics across the tree. Locked in by the integration test."*
- The current behaviour is **safe but loose**: bools silently widen to numeric, but every observed widening site is benign (display formatting, arithmetic with explicit type intent). This issue tightens the language semantics, it does not fix any known correctness bug.

## Notes for Claude

- **Architect first.** Spawn `compiler-architect` to choose between options 1 and 2 before any implementation. The two options have meaningfully different downstream implications (option 2 adds a new kind to the language's mental model; option 1 keeps the kind set minimal).
- The audit phase comes second. Do not start tightening `numeric_type_kind` until you know the audit's volume — if it's >50 sites, the issue likely needs to split into "design + audit" and "execute" sub-issues.
- This issue stays `needs-grooming` until #490 closes. `/pickup` should refuse to claim it before then anyway, but be explicit in the labels.

mcereal:
## Summary

- `numeric_type_kind` now classifies `i1` as its own `"bool"` kind (was `"other"`), so a boolean flowing into a numeric context is a diagnosable ABI mismatch **by construction** rather than a silent `zext`/`uitofp` widening.
- Enforcement at three chokepoints, each emitting a `[fatal]` ABI primitive mismatch with the fix-it ``add `as int` or `as bool` to disambiguate``: the `coerce_operand_to_type` boundary check (call args / struct fields), `dominant_type`/`harmonise_operands` (arithmetic, via the empty-string sentinel), and a new `let`-initializer guard in `instructions_let.sfn` (the `allow_widen=true` path that bypasses the boundary check).
- `as int` is the escape valve. Bool-to-string interpolation is unaffected — it lowers via `select i1` to `"true"`/`"false"`, never numeric coercion.

**Architect decision (recorded on #537): Option 2** — a distinct `"bool"` kind. Under Option 1 (`i1` → `"int"`), `bool → int` looks like `int → int` (same-kind silent widen) and would be *exempted*, so it can't satisfy the negative-test acceptance criterion without extra special-casing. The #490-reserved `i1 → double` bypass now fires, since #490 is closed (per-issue scope decision: tighten everywhere).

Closes #537

## Acceptance Criteria
- [x] Architect decision recorded on the issue (Option 2 — distinct `"bool"` kind).
- [x] `numeric_type_kind("i1") == "bool"`, with diagnostics covering the behaviour.
- [x] Negative test: `let b: bool = true; let n: int = b` fails to compile with the new diagnostic (`compiler/tests/e2e/test_bool_numeric_coercion.sh`).
- [x] Every existing implicit `i1 → numeric` widening in `compiler/src/` and `runtime/prelude.sfn` is migrated or deliberately tolerated — **audit confirmed empirically by self-host**: no compiler-source binding/arithmetic/boundary site relied on bool→numeric widening (else `make compile` would have failed).
- [x] `make compile` self-hosts.
- [x] `make test` passes.
- [x] `make check` green — _running locally to confirm; pushed early to give CI a head start. Will note the result on the PR._
- [x] `sfn fmt --check` clean on every touched `.sfn`.
- [x] Spec update in `site/src/content/docs/docs/reference/spec/06-types.md` §6.

## Verification
```text
make compile  → status: pass (self-hosts; rebuilt after each source edit)
make test     → status: pass; e2e-sh 104/104, 0 failed
test_bool_numeric_coercion.sh → pass=2 fail=0
  • let n: int = b      → refuses, exit 1, "(bool)" + "add `as int` or `as bool`" fix-it
  • let n: int = b as int → compiles, runs, PASS
numeric_abi_mismatch_test.sfn       → 15/15
numeric_int_float_coercion_test.sfn → all pass (bool dominant_type/harmonise cases added)
sfn fmt --check (all touched .sfn)  → rc=0
```

## Files Changed
- `compiler/src/llvm/type_mapping.sfn` — `numeric_type_kind` adds the `"bool"` kind for `i1`.
- `compiler/src/llvm/expression_lowering/native/core_operands.sfn` — boundary `bool → int/float` fatal branch; `dominant_type` symmetric `bool ↔ int/float` sentinel.
- `compiler/src/llvm/lowering/instructions_let.sfn` — `_bool_into_numeric` helper + `let`-initializer guard.
- `compiler/tests/integration/numeric_abi_mismatch_test.sfn`, `compiler/tests/unit/numeric_int_float_coercion_test.sfn` — bool coverage; bypass test flipped to assert refusal.
- `compiler/tests/e2e/test_bool_numeric_coercion.sh` + two fixtures.
- `site/src/content/docs/docs/reference/spec/06-types.md`, `docs/status.md`.

## Scope note (reviewer-flagged follow-up)
Enforcement covers binding initializers, arithmetic, and the call-arg/struct-field boundary (the architect's approved plan). Other `allow_widen=true` value-coercion paths — `return <bool>` from a numeric-returning fn, reassignment to an existing numeric binding, and array/struct element stores — still widen silently and are noted in `docs/status.md` as a tracked follow-up rather than expanding this PR's approved scope.

---
_Generated by [Claude Code](https://claude.ai/code/session_013SHk4Ve5BYUTjgzMnouLVh)_


mcereal:
Thanks @Copilot — good catch, and correct. The fix-it suggested `as bool`, but `as bool` casts are rejected by `lower_cast_expression` (`core_literals_lowering.sfn:1485`, which directs users to a comparison like `x != 0`), so the diagnostic pointed at a non-compiling cast.

Fixed in d9b4db9: the bool→numeric fix-it now reads ``add `as int` or `as float` to disambiguate`` — both are valid for an `i1` source (`i1 as int` → `zext`, `i1 as float` → `uitofp`), matching the existing int↔float diagnostic phrasing. Updated all 8 sites the review flagged: the boundary + `let` diagnostics, the integration/e2e assertions and comments, the fixture comment, the spec §6 note (now explicitly states `as bool` is *not* a valid cast and points at `n != 0`), and the `docs/status.md` row. Re-validated: `make compile` self-hosts, and the bool e2e + integration tests pass against the new wording.

---

Issue #633, opened 2026-05-16
bug(lowering): cross-module `declare` ignores `: int` signatures in `.sfn-asm` interface (declares emit `(double)`)


mcereal:
## Goal

Fix the seed's cross-module signature resolution so that an importer's emitted `declare` line for a function exported from another `.sfn` module respects the **freshly-written `.sfn-asm` interface file** for that module, rather than falling back to a stale or default `(double)` signature.

## Reproducer

Apply this one-character change to `compiler/src/llvm/utils.sfn:131`:

```diff
-fn number_to_string(value: number) -> string {
+fn number_to_string(value: int) -> string {
```

Then:

```bash
ulimit -v 8388608
make clean-build
make compile
```

Observe in `compiler/src/llvm/expressions_helpers.sfn`'s emitted LLVM IR (which imports `number_to_string` from `./utils`):

```llvm
declare i8* @number_to_string__llvm__utils(double)
...
%t5 = call i8* @number_to_string__llvm__utils(double %t4)
```

But:

```bash
$ grep number_to_string build/native/import-context/llvm/utils.sfn-asm
.fn number_to_string(value: int) -> string
```

And the definition in `utils.sfn`'s own emitted IR is:

```llvm
define i8* @number_to_string__llvm__utils(i64 %value) {
```

i.e. the seed wrote an interface file that says `int`, then ignored it and emitted importer-side declares as `double`. The result is hard ABI mismatch between the `i64` definition and every cross-module `(double)` call site. The new binary's `format_temp_name` (which calls `number_to_string`) returns garbage, and `prelude.ll` validation fails with `llvm-as` errors like `multiple definition of local value named 't0'` because the entire SSA-temp naming scheme collapses.

## What works

- **Same-module callers** of a `: int` function are emitted correctly (`(i64 %x)` end-to-end), because the seed resolves the signature locally without going through the interface file.
- **Member-form aliases** that the seed treats as registry-driven (e.g. `runtime.foo`) are not affected — those are governed by `runtime_helpers.sfn` and #572 covers that coupling separately.
- **Pure `: number` chains** continue to work because the legacy default already matches.

## Impact on #611 and the migration

[...]

Issue #306, opened 2026-05-04
compiler: extern call return type defaults to i8* when call result is unused


mcereal:
## Summary

When an imported extern is called as a statement (no consumer for the return value), the call lowering defaults the result type to `i8*` instead of using the imported declaration's actual return type. Produces invalid IR with a call/declare type mismatch.

## Repro

`runtime/sfn/clock.sfn` after the sleep migration:

```sfn
import { sailfin_runtime_sleep } from "./platform/libc";
fn sfn_sleep(seconds: float) -> void ![clock] {
    sailfin_runtime_sleep(seconds);  // statement, return discarded
}
```

`sailfin_runtime_sleep` is declared in `platform/libc.sfn` as `extern fn sailfin_runtime_sleep(seconds: float) -> void;`.

Stage-2 build emits:
```llvm
declare void @sailfin_runtime_sleep(double)
define void @sfn_sleep(double %seconds) {
  %t0 = call i8* @sailfin_runtime_sleep(double %seconds)   ; ← type mismatch
```

The `declare` says `void`, the call site says `i8*`. The compiler defaulted the call's return type because no consumer flowed a constraint back.

For comparison, `runtime/sfn/io.sfn`'s `return write(fd, buf, count);` works correctly because the `return` constraint flows the function's `i64` return type back to the call site.

## Why it matters

- Invalid LLVM IR — clang accepts it for unused returns but the IR is technically malformed.
- Compounds Issue #305 (missing declare line in standalone emit).
- PR 2 of the sleep migration (libc-direct via `nanosleep`) will hit the same problem — `nanosleep`'s return is also typically discarded.
- The architect's plan called this out as Risk #4 / future workaround surface.

## Suggested fix direction

Call-lowering for an imported extern (or any function whose signature is known) should use the declared return type rather than defaulting to `i8*`. The fix likely lives in `compiler/src/llvm/expression_lowering/native/core_call_lowering.sfn` or `core_call_emission.sfn`.

## Workstream context

Discovered during the first C-to-Sailfin runtime migration on `claude/runtime-sfn-sources-sleep-migration`. The migration's call-site sentinel test passes and the binary works because clang accepts the i8*/void mismatch when the result is unused, but PR 2 should not have to rely on this tolerance.

mcereal:
## Architect Review

### Summary

[...]

PR #1134, opened 2026-06-06
feat(runtime): char_from_code raw-byte string write primitive

### LINKED ISSUE #874: feat(runtime): raw-byte string access (char_from_code / byte_at) for pure-Sailfin
## Goal
Provide a raw-byte string primitive (e.g. `char_from_code(n)` and/or `byte_at(s, i)` exposed to capsule code) so pure-Sailfin code can build and hash arbitrary binary/multibyte byte sequences correctly.

## Scope
**In:**
- A prelude-level way to (a) read the raw byte at an index of a string (0–255) and (b) construct a 1-char string from a byte value (`char_from_code`).
- Unit coverage for round-tripping bytes 0–255 and a multibyte UTF-8 string.

**Out:**
- Full `bytes`/`Vec<u8>` type — this is just the minimal byte read/write surface.
- Re-porting SHA-256 to use it (follow-up; #816 shipped with the ASCII-correct `char_code(s[i])` path).

## Acceptance Criteria
- [ ] Raw byte read returns the exact byte (0–255) for every index, including bytes of a multibyte UTF-8 string.
- [ ] `char_from_code(n)` builds the expected 1-byte string for n in 0–255.
- [ ] `make compile` passes (compiler self-hosts).
- [ ] `make check` passes (no regressions).

## Files Affected
- `runtime/prelude.sfn` — expose the byte read + `char_from_code`.
- `compiler/src/llvm/runtime_helpers.sfn` — registry row(s) if a runtime helper backs it.
- `compiler/tests/...` — round-trip coverage.

## Verification
```bash
ulimit -v 8388608 && timeout 60 build/native/sailfin run /tmp/byte_roundtrip.sfn
```

## Context (from #816)
SHA-256 currently decodes input via `char_code(data[i])`, which returns a
codepoint, not a raw byte — correct for ASCII/byte content (all of #816's
vectors) but not for arbitrary binary or multibyte UTF-8 input. There is no
`char_from_code` today: `capsules/sfn/cli/src/style.sfn:5` already carries a
"once char_from_code() support lands" note, and hex output in
`capsules/sfn/crypto/src/mod.sfn` works around its absence with a 16-string
nibble lookup. A real byte primitive removes both workarounds.

## Size
S

## Type
Feature

## Blocked by
None

## Required in pinned seed
None

mcereal:
## Summary
- Add `char_from_code(code: int) -> string` — the raw-byte **write** counterpart to the already-shipped byte read (`char_code(s[i])` returns the raw byte 0–255). Pure-Sailfin code can now both read and construct arbitrary single bytes.
- Writes the byte **verbatim** (no UTF-8 expansion for 128–255), unlike `sfn_str_from_codepoint`, so `char_code(char_from_code(n)) == n` round-trips for every `n` in 1–255.
- Removes the blocker behind the `char_from_code()`-pending workarounds in `capsules/sfn/cli/src/style.sfn` and `capsules/sfn/crypto/src/mod.sfn` (their removal + SHA-256 re-port are follow-ups per the issue `Out:` scope).

Closes #874

## Scope note (resolved with maintainer)
The issue's acceptance asked for round-tripping bytes **0**–255, but byte 0 (NUL) is **unrepresentable** under the current NUL-terminated `string` model — `.length` is recovered with `strlen`, so a lone NUL reads back as length 0. Faithful embedded-NUL storage is a property of the `SfnString` aggregate (explicit length) deferred to **M1.A.2** in `runtime_architecture.md` §2.2, not of this `size:S` helper. This PR ships **1–255** with full fidelity and documents byte 0 (`char_from_code(0) == ""`), asserted in the test so a future representation change that fixes it trips loudly. The redundant read half (`byte_at`) is dropped — the issue's own correction comment confirmed `char_code(s[i])` already returns raw bytes.

## Acceptance Criteria
- [x] Raw byte read returns the exact byte (0–255) for every index, including multibyte UTF-8 — already shipped; pinned by the test (`"é"` → length 2, bytes 195/169).
- [x] `char_from_code(n)` builds the expected 1-byte string for n in **1–255** (byte 0 documented limitation — see scope note).
- [x] `make compile` passes (compiler self-hosts).
- [x] `make check` passes (no regressions) — `e2e-sh: 106/106`, full suite green on both binaries, `stage2 == stage3` fixed point.

## Verification
```text
# new behavioral e2e (build/native/sailfin run on a standalone program)
[test] PASS: char_from_code round-trips 1..255 and writes raw bytes (issue #874)
[summary] 1 passed, 0 failed

# full gate
═══ e2e-sh: 106/106 passed, 0 failed ═══
[check] stage2 == stage3: compiler is a fixed point ✓
status: pass
```

## Files Changed
- `runtime/native/src/sailfin_runtime.c` — `sfn_str_from_byte` C trampoline (malloc'd NUL-terminated 1-byte buffer; structure mirrors `sfn_str_from_codepoint`).
- `runtime/sfn/string.sfn` — `sfn_str_sfn_from_byte` migration proof-of-life wrapper (absorbs the body at M1.A.2).
- `runtime/prelude.sfn` — `char_from_code`, calling the C symbol via an inline `extern fn` (the #306 cross-module workaround `exec.sfn` uses — prelude emission runs before import contexts are staged, so an imported wrapper can't resolve its signature under the seed) + alloca-backed `* u8 as string` cast.
- `compiler/src/llvm/runtime_helpers.sfn` — bare-name `char_from_code` descriptor so user call sites resolve (same shape as the `char_code` row); self-host is unaffected because the prelude's own body resolves through its function entry (#633), not the descriptor.
- `compiler/tests/e2e/test_runtime_char_from_code.sh` — 1–255 round-trip, raw high byte 200 stays one byte, multibyte read intact, byte-0 empty-string case.
- `docs/status.md` + `site/.../reference/standard-library.md` — status row + API reference.

## Notes for future grooming
Exposing a *new* runtime-backed prelude function hit two self-host traps worth recording: (1) the seed can't resolve a cross-module imported wrapper's signature during prelude emission (#306) → use an inline `extern fn`; (2) user programs can't resolve the call's return type without a bare-name runtime-helper descriptor. Both are now handled, but a "Required in pinned seed: None" issue of this shape should expect them.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XTCJ9taQ4P2iP9qfh9MtCK)_


mcereal:
macOS-arm64 CI (`e2e-sh-1`) caught a pre-existing `char_code` quirk that the full-range round-trip surfaced — worth recording here.

`grapheme_at` encodes an ASCII byte as the immediate-codepoint tagged pointer `(byte << 32)`. For byte `1` that's address `0x100000000`, which is **exactly the arm64 executable load base**, so `_is_immediate_codepoint_string`'s `mach_vm_region` guard sees a mapped region, treats the tag as a real pointer, and `char_code` dereferences the Mach-O header (`0xCF`) instead of decoding `1`. `char_from_code(1)` itself writes the byte correctly — the defect is on the **read** side, and it affects any string's `0x01` byte on macOS arm64 (just rare in normal text, which is why it was never caught).

Per the scope-out decision: narrowed the round-trip loop to `2..255`, added a write-side length assertion for byte 1, and documented the caveat. The real fix to the immediate encoding is tracked in #1136. Linux and bytes ≠ `0x01` are unaffected.

---

Issue #1146, opened 2026-06-06
feat(lowering): materialize defined-fn address as i8* in value position


mcereal:
## Goal
A defined (or extern) function name in value position lowers to its bare `@symbol` bitcast to `i8*`, so `worker as * u8` (and a typed C-ABI fn-pointer parameter) receives a real code pointer instead of silently lowering to `null`.

## Scope
**In:**
- In `lower_expression` (`compiler/src/llvm/expression_lowering/native/core.sfn`), after the param lookup (~line 1558) and **before** the fallthrough diagnostic (~line 1564): look up the stripped name via `find_function_by_name_or_import(functions, ...)` (`rendering_helpers.sfn:361`). On a hit, emit a value of LLVM type `i8*` that is a `bitcast <fnty> @<NativeFunction.name> to i8*` — a **single `i8*`**, never the `{i8*, i8*}` closure pair.
- Reconstruct the fn-pointer LLVM type from the `NativeFunction` record (reuse the lambda fn-ptr-type helper at `core.sfn:~698`).
- Use the bitcast instruction template from `closures.sfn:544`, but emit a single `i8*` (do **not** engage the `__closure__` sentinel / pair path at `closures.sfn:512-513`).
- Extern functions: resolve `extern fn` declarations through the same path so `some_c_entry as * u8` → `@some_c_entry`. If extern decls are tracked separately, resolve against that list too.
- Primary spelling is `<fn> as * u8` (the cast value path delegates to `lower_expression`; once it returns a real `i8*`, the `i8*`==`i8*` passthrough at `core_literals_lowering.sfn:1463` returns it directly — **no cast-path change needed**).
- New e2e link round-trip test + an IR-shape unit test.

**Out:**
- The `& fn` spelling (defer to post-first-class-functions era — `&` means `&raw <local>` today).
- Bare `worker` (no cast) arg-type inference (entangles with #1118 call dispatch).
- The error diagnostic for unsupported/un-lowerable references (that is the follow-up issue).
- Cross-module fn references if found broken (conditional follow-up).
- Closure/indirect **apply** / invoking a fn-pointer value (epic #1118).

## Acceptance Criteria
- [ ] `emit llvm` on `let s: * u8 = worker as * u8;` produces `bitcast <fnty> @worker... to i8*` and stores the **non-null** result (assert IR contains `bitcast` + `@worker`, and contains **no** `store i8* null` for that slot).
[...]

PR #1151, opened 2026-06-07
feat(typecheck): error on unsupported function references in value position (E0808/E0809)

### LINKED ISSUE #1147: feat(typecheck): error on un-lowerable function references in value position
## Goal
A function reference that #1146 cannot lower to a bare code pointer (generic fn, non-C-ABI signature, or an unsupported spelling like `& fn` / bare-arg) produces a clear compile-time **error** with a source span — never a silent `null` or an elided call.

## Scope
**In:**
- New typecheck error code (next free in the E080x extern family — verify highest currently used; I saw E0807 at `typecheck_types.sfn:1516`, so likely **E0808**) in `compiler/src/typecheck_types.sfn`, fired when a bare identifier resolving to a function appears in value position and the target/result is not the supported `* u8` code-pointer form. Message names the function and points at the supported form (`<fn> as * u8`).
- Upgrade the emit-side fallthrough diagnostic at `core.sfn:~1564` to **`[fatal]`** as a backstop, so an unresolved value-position function reference fails the build instead of eliding the enclosing call.

**Out:**
- Any happy-path lowering (that is #1146).
- Broadening the accepted set of referenceable signatures.

## Acceptance Criteria
- [ ] A generic or non-C-ABI fn used as `<fn> as * u8` reports the new error with a source span; `sfn check` exits non-zero.
- [ ] `& worker` (still unsupported) reports a clear error, not a silent null.
- [ ] #1146's happy-path cases still emit correctly (run its IR-shape test unchanged as a regression guard).
- [ ] No emitted IR ever contains a `store i8* null` arising from a dropped function reference (the silent-null class is eliminated).
- [ ] `make compile` passes (self-hosts against pinned seed)
- [ ] `make test` passes
- [ ] `sfn fmt --check` clean on every touched .sfn

## Files Affected
- compiler/src/typecheck_types.sfn — new E0808 + diagnostic constructor near the E080x family (~1175-1256); the value-position fn-ref check where idents are type-resolved
- compiler/src/llvm/expression_lowering/native/core.sfn — upgrade the :1564 fallthrough to fatal
- compiler/tests/unit/ — negative typecheck tests
- docs/status.md + relevant spec §/preview entry for the new error code

## Verification
```bash
ulimit -v 8388608 && make compile && make test
# focused: a generic-fn / & fn reference must produce the new E08xx diagnostic, not silent null
```

## Size
S

## Type
Feature

## Blocked by
#1146

## Required in pinned seed
None

## Context
The safety half of the #1142 capability — closes the silent-null class that "let #1088 slip through grooming" (typecheck accepted a function reference that emit then dropped to `null`). Must be ordered **after** #1146 so the guard lets the now-supported forms through. Groomed from #1142 on 2026-06-06. Seed-completeness gate for #1088: the seed #1088 builds against should contain both #1146 and this issue.

mcereal:
## Summary
- Adds **E0808** (a function used as a value in an unsupported position — bare name, `& fn`, or `<fn> as <non-pointer>`) and **E0809** (`<fn> as * u8` where the function is generic or non-C-ABI) to the typecheck walker.
- Together with #1146 (which lowers the supported `<fn> as * u8` form), this closes the silent-`null`-from-a-dropped-function-reference class **at the frontend** — no function reference reaches lowering to be dropped.
- The parser degrades both `& fn` and `<fn> as *T` to `Raw` text, so detection lives in a new `Identifier` arm + a `Raw` arm, gated on the head resolving to a `function`/`extern function` symbol.

Closes #1147

## Design notes
- **Parser reality:** `<fn> as *T` and `& fn` don't parse as structured `Cast`/`Unary` nodes (the `as` target only accepts a bare identifier; `&` has no prefix form) — both fall back to `Expression.Raw`. So the checks are a `Raw`-text classifier (`check_fn_reference_raw`) plus a bare-`Identifier` arm, not a structured `Cast` arm.
- **Self-host gate:** the diagnostics fire **only** when the head identifier resolves to a function/extern-function symbol, so non-function `<value> as * u8` casts (the driver's `label as * u8`, `argv[0] as * u8`) stay clean.
- **Signature classification:** `SymbolEntry` gains `is_generic`/`is_c_abi`, populated from the signature at registration (`signature_is_generic`/`signature_is_c_abi`, threaded via defaulted params so existing `register_*` callers are untouched) — the walker classifies references without re-deriving signatures or threading the program AST.
- **No false-positives on receiver positions:** the `Call` arm doesn't treat a bare-identifier callee as a value, and the `Member` arm doesn't treat a bare-identifier object as a value (`env.get(...)`, `Color.Red`).
- The blanket "fatal fallthrough" hardening explored during review was **dropped**: it forced pre-existing, unrelated silent-drops of non-function globals (`EXIT_OK`, `EXIT_USAGE`, `greeting`) to hard-fail. AC#4 is scoped to *function references*, which are now all caught at typecheck — so the emitter backstop is unnecessary and out of scope.

## Acceptance Criteria
- [x] Generic / non-C-ABI fn used as `<fn> as * u8` reports an error with a source span; `sfn check` exits non-zero (E0809)
- [x] `& worker` (and bare-name / `as <non-pointer>`) reports a clear error, not a silent null (E0808)
- [x] #1146's happy-path cases still emit correctly (`test_fn_reference_pthread.sh` green)
- [x] No emitted IR contains a `store i8* null` arising from a dropped function reference (all caught at typecheck before lowering)
- [x] `make compile` self-hosts; `make test` passes; `sfn fmt --check` clean

## Verification
```text
ulimit -v 8388608 && make compile          # self-hosts (exit 0)
build/native/sailfin check compiler/src/ runtime/   # 176 files ok, 0 false-positives
build/native/sailfin check capsules/                # 66 files ok, 0 false-positives
make test:
  unit 132/132 · integration 31/31 · e2e 5/5 · capsules 34/34 · e2e-sh 110/110   (exit 0)
compiler/tests/unit/fn_reference_typecheck_test.sfn   # 9/9
compiler/tests/e2e/test_fn_reference_pthread.sh        # 4/4 (#1146 regression guard)
```

## Files Changed
- `compiler/src/typecheck.sfn` — `Identifier`/`Raw` walker arms; `Call`/`Member` receiver guards; flag-populating registration
- `compiler/src/typecheck_types.sfn` — `SymbolEntry.is_generic`/`is_c_abi`; `find_symbol`, `is_function_kind`, `signature_is_*`, `check_fn_reference_raw`, E0808/E0809 factories
- `compiler/tests/unit/fn_reference_typecheck_test.sfn` — 9 cases (new)
- `docs/status.md` — E0808/E0809 entry

---
_Generated by [Claude Code](https://claude.ai/code/session_01RjSNQawEfpXygmAzgKMYPg)_

### DIFF
--- a/compiler/src/ast.sfn
+++ b/compiler/src/ast.sfn
@@ -124,7 +124,7 @@ enum Expression {
     // `serve(handler, port)` — accept loop dispatching to the thread pool;
     // `port` is optional to allow a configured default (§2.6.6).
     Serve { handler: Expression, port: Expression?, span: SourceSpan? },
-    Raw { text: string }
+    Raw { text: string, span: SourceSpan? }
 }
 
 struct Parameter {
--- a/compiler/src/emit_native_desugar_try.sfn
+++ b/compiler/src/emit_native_desugar_try.sfn
@@ -752,12 +752,12 @@ fn dt_build_match(statement: Statement, rest: Statement[], id: int) -> DtStmt {
     };
 
     let ok_case = MatchCase {
-        pattern: Expression.Raw { text: "Result.Ok { value }" },
+        pattern: Expression.Raw { text: "Result.Ok { value }", span: null },
         guard: null,
         body: ok_result.block
     };
     let err_case = MatchCase {
-        pattern: Expression.Raw { text: "Result.Err { error }" },
+        pattern: Expression.Raw { text: "Result.Err { error }", span: null },
         guard: null,
         body: err_block
     };
--- a/compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn
+++ b/compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn
@@ -342,7 +342,7 @@ fn _prepend_capture_prologue(body: Block, lambda_id: int, captures: Capture[]) -
     if captures.length == 0 { return body; }
     let marker_text = closure_env_load_prologue_marker(lambda_id, _encode_captures(captures));
     let prologue_stmt = Statement.ExpressionStatement {
-        expression: Expression.Raw { text: marker_text },
+        expression: Expression.Raw { text: marker_text, span: null },
         span: null
     };
     let mut new_statements: Statement[] = [prologue_stmt];
@@ -572,7 +572,7 @@ fn _rewrite_lambda(ctx: LiftContext, lambda: Expression) -> ExpressionRewriteRes
         let marker_text = closure_pair_marker(lambda_id, "null");
         return ExpressionRewriteResult {
             ctx: new_ctx,
-            expression: Expression.Raw { text: marker_text }
+            expression: Expression.Raw { text: marker_text, span: null }
         };
     }
 
@@ -585,7 +585,7 @@ fn _rewrite_lambda(ctx: LiftContext, lambda: Expression) -> ExpressionRewriteRes
         let fatal_text = closure_pair_fatal_mutable_marker(ctx.counter, mutable_name);
         return ExpressionRewriteResult {
             ctx: ctx,
-            expression: Expression.Raw { text: fatal_text }
+            expression: Expression.Raw { text: fatal_text, span: null }
         };
     }
 
@@ -601,7 +601,7 @@ fn _rewrite_lambda(ctx: LiftContext, lambda: Expression) -> ExpressionRewriteRes
     let marker_text = closure_pair_env_alloc_marker(lambda_id, _encode_captures(record.captures));
     return ExpressionRewriteResult {
         ctx: new_ctx,
-        expression: Expression.Raw { text: marker_text }
+        expression: Expression.Raw { text: marker_text, span: null }
     };
 }
 
--- a/compiler/src/parser/expressions.sfn
+++ b/compiler/src/parser/expressions.sfn
@@ -70,7 +70,7 @@ fn expression_tokens_advance(state: ExpressionTokens) -> ExpressionTokens {
 fn expression_parse_failure(state: ExpressionTokens) -> ExpressionParseResult {
     return ExpressionParseResult {
         state: state,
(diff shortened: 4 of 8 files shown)


mcereal:
Addressed all five review comments in `b55b35a`:

1. **`docs/status.md` heading** — dropped the stale "+ fatal silent-drop backstop" from the title (the backstop was removed; the body already said so).
2. **Unused `Diagnostic` import** — removed from `fn_reference_typecheck_test.sfn`.
3. **Any `*T` target accepted** — tightened: `check_fn_reference_raw` now treats **only `* u8`** as the supported address form. A non-`u8` pointer target (`worker as * i32`, which would reinterpret a code address as a typed data pointer) is now **E0808**, matching the documented contract.
4. **Parenthesized `(worker) as * u8` slipping past** — added `_strip_outer_parens`; `(gen) as * u8` is now correctly **E0809** and `(worker) as * u8` is correctly accepted (new regression tests for both).
5. **Raw forms had no source span** — `Raw` AST nodes now carry a `span`, populated at the parser's degradation sites via `source_span_from_tokens`. `& fn` / `<fn> as *T` diagnostics now caret the source location, e.g.:
   ```
   error[E0809]: cannot take the address of function `gen` because it is generic; ...
     --> file.sfn:4:13
      |
    4 |     let f = gen as * u8;
      |             ^^^
   ```

Validation: `make rebuild` self-hosts; new-binary audit clean across **242** compiler+runtime+capsule files (0 false-positives); unit suite extended to **12/12**; integration 31/31, e2e 5/5, capsules 34/34, e2e-sh 110/110.

Note on the full `make test` run: one unit shard reported a failure from the **known SIGSEGV-139 flake** (`method_target_unresolved_struct_test` / `struct_field_pointer_assignment_test` — unrelated to this change; the harness has a built-in retry-once for exit 139). Both **pass in isolation**. CI re-runs fresh.

---

PR #1175, opened 2026-06-07
fix(lowering): multi-capture lambda env struct missing fields (invalid GEP)

### LINKED ISSUE #1106: fix(lowering): two-capture lambda emits env struct missing its second field (invalid GEP)
## Goal
A lambda that captures **two or more** enclosing locals miscompiles: the `%sfn_closure_env_<id>` struct type is emitted with only the *first* capture's field, while the env-load prologue inside the lifted function GEPs the second field — producing invalid LLVM IR that fails at link. Single-capture lambdas are correct. Make multi-capture emit the full struct.

## Reproduction
Minimal, no I/O, against `sfn 0.7.0-alpha.25` self-hosted (`build/native/sailfin`):

```sfn
// two-int capture
fn apply(cb: fn(int) -> int, x: int) -> int { return cb(x); }
fn main() ![io] {
    let a: int = 10;
    let b: int = 20;
    let r = apply(fn(d: int) -> int { return d + a + b; }, 1);
    print(number_to_string(r));
}
```

```
$ ulimit -v 8388608 && build/native/sailfin run /tmp/cap2i.sfn
build/sailfin/run.ll:227:52: error: invalid getelementptr indices
  %t3 = getelementptr inbounds %sfn_closure_env_0, %sfn_closure_env_0* %t0, i32 0, i32 1
link failed (clang exit=1)
```

Two-string capture (`let a; let b; ... d + a + b`) fails identically. **One** capture is fine. Confirmed still reproducing on `main` @ self-hosted `build/native/sailfin` (2026-06-07).

## Scope
**In:**
- Make the env-struct **type-decl**, the **alloc/populate** walk, and the **load prologue** agree on field count + order for N≥2 captures.
- Reconcile the two marker encode sites that drift: the alloc/type-decl side feeds `_encode_captures(record.captures)` (`lambda_lowering.sfn:601` → consumed in `core.sfn:_lower_closure_env_alloc`), while the load-prologue side feeds `_encode_captures(captures)` (`lambda_lowering.sfn:343` → consumed in `instructions.sfn:_lower_closure_env_load_prologue_marker`). Both decode via `decode_captures` + `synthesize_closure_env_struct`, yet the emitted IR shows 1 field on the alloc/type-decl side and 2 on the prologue side — establish a single source of truth for the capture set so all three emission sites cannot diverge.
- Add a **runtime e2e** fixture (2- and 3-capture lambdas) under `compiler/tests/e2e/fixtures/`.

**Out:**
- Mutable-capture semantics (still refused by `_find_mutable_capture_name`; by-value snapshot only — unchanged).
- Closure ABI redesign / the dead-code `closures.sfn` helper API surface beyond what the live path imports.
- `array.map`/`filter`/`reduce` over closures (#399) — this fix is a *prerequisite*, not part of it.

## Root cause (from the emitted IR)
```llvm
%sfn_closure_env_0 = type { i64 }            ; <-- only ONE field for a TWO-capture lambda

; alloc/populate site: only field 0 is stored
%t5 = getelementptr inbounds %sfn_closure_env_0, %sfn_closure_env_0* %t4, i32 0, i32 0

; load prologue inside the lifted fn: reads field 0 AND field 1
%t1 = getelementptr inbounds %sfn_closure_env_0, %sfn_closure_env_0* %t0, i32 0, i32 0
%t3 = getelementptr inbounds %sfn_closure_env_0, %sfn_closure_env_0* %t0, i32 0, i32 1   ; field 1 doesn't exist
```

The env-struct **type declaration** (emitted in the alloc path, `core.sfn:550` from `synthesize_closure_env_struct`) and the **alloc/populate** sequence agree on a 1-field struct, while the **load prologue** expects 2. The alloc/type-decl marker and the prologue marker are fed capture sets from different sources (`record.captures` vs the `captures` threaded into the lifted-function prologue), so the field counts drift. The three sites must derive their field list from one place.

## Why this slipped through
#459/#687/#689 only verified multi-capture with **golden IR** integration tests (`closure_env_emission_test.sfn`) and deterministic capture *ordering* — never a runtime e2e. The only runnable closure e2e fixtures (`compiler/tests/e2e/fixtures/closure_single_capture`, `closure_higher_order`) cover **single** capture and **zero**-capture HOF. So a multi-capture program has never actually been compiled-and-run in CI.

## Acceptance Criteria
- [ ] A 2-capture lambda (two ints; two strings; mixed int+string) compiles, links, runs, and prints the correct result.
- [ ] The `%sfn_closure_env_<id>` type decl, the alloc/populate sequence, and the load prologue all agree on field count and order for N≥2 captures.
- [ ] New **runtime e2e** fixture under `compiler/tests/e2e/fixtures/` exercising a 2- and 3-capture lambda (golden IR alone is insufficient — that's how this slipped).
- [ ] `make compile` self-hosts; `make check` green.
- [ ] `sfn fmt --check` clean on every touched `.sfn`.

## Files Affected (initial map)
- `compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn` — the two `_encode_captures` call sites (`:343` prologue, `:601` alloc) that drift on capture source.
- `compiler/src/llvm/expression_lowering/native/core.sfn` — `_lower_closure_env_alloc` (`:537`): type-decl + alloc walk.
- `compiler/src/llvm/lowering/instructions.sfn` — `_lower_closure_env_load_prologue_marker` (`:668`): load-prologue field walk.
- `compiler/src/llvm/closures.sfn` — shared `synthesize_closure_env_struct` / `decode_captures` / `emit_closure_env_*` helpers (verify they aren't the truncation point; they currently walk all fields).
- `compiler/tests/e2e/fixtures/` + `compiler/tests/e2e/test_closure_capture.sh` — add multi-capture runtime cases.

## Verification
```bash
# Repro fixture (must print 31, then exit 0)
cat > /tmp/cap2i.sfn <<'SFN'
fn apply(cb: fn(int) -> int, x: int) -> int { return cb(x); }
fn main() ![io] {
    let a: int = 10;
    let b: int = 20;
    let r = apply(fn(d: int) -> int { return d + a + b; }, 1);
    print(number_to_string(r));
}
SFN
ulimit -v 8388608 && timeout 60 build/native/sailfin run /tmp/cap2i.sfn   # expect: 31

# Self-host + suite
make compile
make check
ulimit -v 8388608 && build/native/sailfin fmt --check compiler/src/ runtime/
```

## Size
S

## Type
Bug

## Blocked by
None

## Required in pinned seed
None

## Context
Surfaced while implementing #979 (scoped `with_tmp_dir` test fixture): a fixture-test lambda that captured two strings (a `token` + a marker path) failed to compile with the GEP error above. #979 shipped by constraining its test lambdas to a single capture each; this bug tracks the underlying fix.

Tagged `seed-blocker`: multi-capture is a prerequisite for `array.map`/`filter`/`reduce` over closures (#399) and any spawn handler that closes over >1 value — those downstream consumers can only self-host once this fix is in the pinned seed, so it must close before the next seed bump.

mcereal:
## Summary
- A lambda capturing **two or more** enclosing locals miscompiled: the `%sfn_closure_env_*` struct was emitted with only the first capture's field while the lifted body's load prologue GEP'd the second field — invalid LLVM IR that failed at link (`invalid getelementptr indices`). Single-capture lambdas were fine.
- **Root cause:** the env-alloc marker's capture list was comma-separated (`captures=a:int,b:int`). When a capturing lambda is passed inline as a call argument (`apply(fn … , 1)`), the comma-splitting argument parser `split_array_elements` (tracks `()`, `[]`, `{}`, strings — but **not** `<>`) split the marker at the comma between captures, truncating the first argument to `captures=a:int`. The alloc/type-decl side then saw one capture while the standalone-`eval` load prologue saw both. *Note:* the issue's "two encode sites drift" diagnosis was slightly off — both markers are byte-identical; the divergence is purely the call-arg splitter eating the separator.
- **Fix:** change the inter-capture separator to `;` in the shared `_encode_captures`/`decode_captures` pair — the true single source of truth for the capture set. No argument splitter splits on `;`, so the marker survives transport and all three emission sites (type decl, alloc/populate, load prologue) derive the same field list. Making the splitter `<>`-aware was rejected — it would break comparison operators (`a < b`) and generics in argument position.
- Adds **runtime e2e** fixtures (two-int, three-int, mixed string+int) wired into `test_closure_capture.sh` — golden IR alone is how this slipped (no multi-capture program had ever been compiled-and-run in CI).

Closes #1106

## Acceptance Criteria
- [x] A 2-capture lambda (two ints; mixed int+string) compiles, links, runs, and prints the correct result. *(also covers two-string-style via the mixed fixture)*
- [x] The `%sfn_closure_env_` type decl, the alloc/populate sequence, and the load prologue all agree on field count and order for N≥2 captures.
- [x] New **runtime e2e** fixture exercising 2- and 3-capture lambdas under `compiler/tests/e2e/fixtures/`.
- [x] `make compile` self-hosts.
- [x] `sfn fmt --check` clean on every touched `.sfn`.
- [ ] `make check` green — re-running locally; CI is the authoritative gate. The first local run failed only on two **transient** segfaults in unrelated tests (`method_target_unresolved_struct_test`, `struct_field_pointer_assignment_test`), both of which pass deterministically in isolation; all closure tests (integration 31/31, e2e-sh 111/111 including the new fixtures) passed.

## Verification
```text
$ build/native/sailfin run /tmp/cap2i.sfn        → 31   (was: link failed, invalid GEP)
$ grep 'sfn_closure_env_0 = type' build/sailfin/run.ll
  %sfn_closure_env_0 = type { i64, i64 }         (was: { i64 })

$ bash compiler/tests/e2e/test_closure_capture.sh build/native/sailfin
  PASS (1) canonical lambda-closure prints 7
  PASS (2) single-int-capture prints 12
  PASS (3) higher-order fn(int)->int dispatch prints 10
  PASS (4) two-int-capture prints 31
  PASS (5) three-int-capture prints 16
  PASS (6) mixed string+int two-capture prints xv42
  6 passed, 0 failed

$ build/native/sailfin test compiler/tests/integration/closure_{lifting,env_emission}_test.sfn  → all pass
$ make compile  → self-hosts (exit 0)
$ build/native/sailfin fmt --check compiler/src/ runtime/  → clean
```

## Files Changed
- `compiler/src/llvm/closures.sfn` — `_encode`/`decode` separator `;`, escape `;`→`%3B`
- `compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn` — `_encode_captures` joins with `;`
- `compiler/src/llvm/expression_lowering/native/core.sfn` — marker doc comment
- `compiler/tests/integration/closure_lifting_test.sfn` — golden assertions + `;` round-trip case
- `compiler/tests/e2e/fixtures/closure_{two_int,three_int,mixed}_capture/` + `test_closure_capture.sh` — new runtime e2e
- `docs/status.md` — note the #1106 multi-capture fix

---
_Generated by [Claude Code](https://claude.ai/code/session_01K5LrJSWm4GebGaRRhQ49Mi)_


mcereal:
Addressed the Copilot review in `16bfb7b`. All four comments flag the same real latent issue: the `captures=` value is extracted by `_parse_marker_string_field` / `_parse_prologue_string`, which terminate at the next space, so a capture whose `type_text` contains a space (`Map<string, int>`, `fn(int) -> boolean`) would truncate the encoded list and drop later captures — the same missing-env-fields bug class this PR fixes for the comma case.

Fix: `percent_escape_capture_text` now escapes ASCII space to `%20` (and `percent_unescape_capture_text` decodes it back), so the encoded capture list is whitespace-free and survives the space-terminating field parsers. Space-free segments (the common int/string/bool case) are byte-identical. Updated the golden round-trip + decode tests to the `%20` form and added an explicit space round-trip case.

Verified locally on a clean `make clean-build` + `make compile` (self-hosts): repro prints `31`, all six `test_closure_capture.sh` cases pass, and `closure_lifting_test.sfn` (15 tests, incl. the `%20` decode case) passes. CI is re-running.

---

PR #1231, opened 2026-06-09
perf(test-runner): content-addressed per-test binary cache (#1230)

### LINKED ISSUE #1230: perf(test-runner): content-addressed per-test binary cache (skip re-lower+relink on unchanged tests)
**Parent epic:** #843 (Phase 4 — Parallel test execution) · umbrella #839 · **Track A, Lever 2** of `docs/proposals/ci-test-speed.md`.

## Goal
Make `make test` incrementally fast by content-addressing each test's linked native binary, so an unchanged test skips LLVM lowering + the clang link (the dominant per-test cost) and just runs the cached binary.

## Context / why
The runner (`handle_test_command`, `compiler/src/cli_commands.sfn:958`) forks one `sfn test <file>` child per test file in a serial loop, and **every run** re-lowers each test's full transitive `compiler/src` closure and **clang-links** it from scratch — there is *no per-test binary cache*. That redundant lower+link across ~145 unit binaries is the bulk of the ~22 min unit suite (the module IR cache in `build/cache/` already makes the compiler *build* ~free on a warm cache, but it does **not** cover test-binary lowering/linking). CI sharding (#1012, shipped) only parallelizes CI wall-time — it reduces zero total work and does nothing for local `make test`. This issue is the fundamental incremental fix; full design in `ci-test-speed.md` §"Lever 2 — Per-test binary cache" (variant **2a**, the conservative "cache the binary, still run it" form).

## Scope
**In:**
- Derive a per-test-binary cache key reusing the existing `compiler/src/build_cache.sfn` primitives (`cache_key_for`/`_sha256_of_string`/`canonicalize_flags`/`build_cache_schema_version`): `sha256(content_hash(test.sfn) || sorted(hash of each transitive dep the link consumes) || compiler_version || canonical(clang flags) || schema_version)`. **Derive the dep set from the same resolver output the link already uses** — `TestGroup.native_texts` + `TestGroup.dep_ll_paths` (`cli_commands.sfn:362-367`, populated at `:1440-1441`) — so the key can never miss a dep (the correctness hinge: a changed dep → changed hash → cache miss, same model as `cache_key_for`).
- Store the linked test executable under `build/cache/test-bin/v1/<key>` with an **atomic** write (write to a unique temp in scratch, then rename into place — reuse the `_sha256_of_string` unique-suffix / hash-tmp precedent) so racing shards never half-write.
- In the per-group compile→link path (`cli_commands.sfn:~1553` lower, `~1588` `_clang_link_test_cmd_with_deps`): on a cache **hit**, skip the lower + link and run the cached binary; on a **miss**, lower+link as today then populate the cache. **Always run the binary** (variant 2a — never trust a cached pass/fail result).
- Add a `--no-test-cache` flag to `handle_test_command` that bypasses both read and write, and wire `make check` / the nightly full-suite path to pass it, so the merge/nightly gate always cold-builds test binaries and the cache can never mask a test-compile regression.
- Emit `cache.test_bin_hit_rate` in the `sfn test --json` run summary (mirror the module-cache hit-rate field already in the build `--json` summary).
- Regression test: a unit test pinning that the key changes when a dep hash changes and is stable for identical inputs.

**Out:**
- Variant **2b** (cache the pass/fail *result* and skip the run; `SAILFIN_TEST_RESULT_CACHE=1`) — separate follow-up, higher risk.
- CI save/restore plumbing for `build/cache/test-bin` (`.github/actions/sailfin-build`, nightly baseline warm) — separate CI-config issue (Lever 2 "CI plumbing").
- `sailfin test --shard I/N` deterministic partition (Lever 1 Seam B).
- In-process `--jobs N` parallelism (Track B — post-1.0, blocked on `routine`/`spawn`).

## Acceptance Criteria
- [ ] A no-op rerun of a suite reports ~100% `cache.test_bin_hit_rate` in `--json`, and is materially faster than the cold run (lower+link skipped for unchanged tests).
- [ ] After editing one `compiler/src/*.sfn`, the next run shows cache **misses** for exactly the tests that transitively import it and **hits** for the rest (no false hits, no false misses).
- [ ] `--no-test-cache` forces a cold lower+link (hit rate 0); `make check` passes it so the full-suite gate never reads the cache.
- [ ] Cache writes are atomic (temp + rename); two concurrent shards on the same key cannot corrupt it.
- [ ] `make compile` self-hosts; `make test` passes (no regressions); `sfn fmt --check` clean on touched `.sfn`.

## Files Affected
- `compiler/src/cli_commands.sfn` — per-test-binary cache key derivation; hit-check before lower+link and populate-after in the per-group loop (`~1553`/`~1588`); `--no-test-cache` flag parse in `handle_test_command`; `cache.test_bin_hit_rate` in the `--json` summary.
- `compiler/src/build_cache.sfn` — reuse/extend cache-key + atomic-write helpers; a `test-bin/v1` path helper if `cache_artifact_path` doesn't already cover it.
- `Makefile` — pass `--no-test-cache` on the `make check` / full-suite gate path.
- `compiler/tests/unit/<new>_test.sfn` — cache-key derivation regression test.

## Verification
```bash
ulimit -v 8388608 && make compile
# cold run, then identical rerun should be hit-heavy + faster:
ulimit -v 8388608 && build/native/sailfin test compiler/tests/unit --json | tail -5   # baseline
ulimit -v 8388608 && build/native/sailfin test compiler/tests/unit --json | tail -5   # ~100% test_bin_hit_rate
# edit one compiler/src dep, rerun: misses only for transitive importers
ulimit -v 8388608 && build/native/sailfin test compiler/tests/unit --no-test-cache --json | tail -5  # hit rate 0
ulimit -v 8388608 && make test
```

## Size
M

## Type
Performance

## Blocked by
None — Track A needs no concurrency runtime; builds on the shipped module-cache machinery.

## Required in pinned seed
None — runner-only change building on `compiler/src/build_cache.sfn` primitives already in-tree; no new frontend/backend primitive required.

mcereal:
## Summary

- Content-addresses each test's linked native binary so an **unchanged** test skips LLVM lowering + the clang link (the dominant per-test cost) and just re-runs the cached executable. Lever 2 (variant 2a) of `docs/proposals/ci-test-speed.md`.
- On a hit the binary is still **run** (never a cached pass/fail result); on a miss it cold lower+links then populates the cache.
- `--no-test-cache` bypasses read + write; `make check` passes it so the merge/seedcheck gate always cold-builds and a test-compile regression can never be masked by a stale hit.

Closes #1230

## What changed

- **`compiler/src/build_cache.sfn`** — `test_bin_cache_key` folds `sha256(test source)` + the resolver's transitive dep closure (`TestGroup.native_texts` + `dep_ll_paths` — the *exact* link inputs, so the key can't miss a dep) + `cache_compiler_identity` (a rebuilt compiler busts every entry) + canonical clang flags + a `test-bin/v1` schema. Binaries store under `build/cache/test-bin/v1/` via the existing **atomic temp+rename** artifact helpers (new `"test-bin"` kind).
- **`compiler/src/cli_commands.sfn`** — `--no-test-cache` flag parse; per-test lookup→hit (skip lower+link, re-assert `+x` on the restored binary — the shared `mktemp` copy strips the exec bit) / miss (lower+link then store); hit/miss accounting. Per-file children forward the flag; the multi-file parent aggregates `test_bin_hits`/`misses` from each child's sub-frame summary.
- **`compiler/src/test_runner_json.sfn`** — additive `cache` object in the `--json` summary (`test_bin_hits`/`misses` + derived `test_bin_hit_rate`, 4-decimal floor, mirroring the build report's `cache.hit_rate`). No schema bump (additive optional field).
- **`Makefile`** — `TEST_BIN_CACHE_FLAGS`; `make check` passes `--no-test-cache` on both gate passes.
- Docs: spec §11 (`11-testing.md`), `docs/status.md`, proposal status row.

## Acceptance criteria

- [x] No-op suite rerun reports ~100% `cache.test_bin_hit_rate` and is materially faster — measured **integration suite 161s cold → 53s warm (~3×)**; trivial single test 326ms→29ms.
- [x] Editing a dep content-busts the key → miss for transitive importers, hit for the rest (no false hits/misses) — pinned by unit tests + e2e edit case.
- [x] `--no-test-cache` forces a cold lower+link (hit rate 0); `make check` passes it.
- [x] Cache writes are atomic (temp + rename) — reuses the shared `_cache_atomic_copy` helper.
- [x] `make compile` self-hosts; warm `make test` passes (269/269 integration, full suite clean); `sfn fmt --check` clean on touched `.sfn`.

## Verification

```
# cold then warm (isolated cache):
build/native/sailfin test compiler/tests/integration --json | jq .cache
#  cold → {"test_bin_hits":0,"test_bin_misses":31,"test_bin_hit_rate":0.0000}  (161s)
#  warm → {"test_bin_hits":31,"test_bin_misses":0,"test_bin_hit_rate":1.0000}  ( 53s)
build/native/sailfin test compiler/tests/integration --no-test-cache --json | jq .cache
#       → {"test_bin_hits":0,"test_bin_misses":0,"test_bin_hit_rate":0.0000}
bash compiler/tests/e2e/test_test_bin_cache.sh build/native/sailfin   # 4/4 passed
build/native/sailfin test compiler/tests/unit/build_cache_test.sfn    # 63/63 (incl. 7 new)
```

## Notes / scope

- **Incremental-only by design.** A *cold* run (CI, fresh checkout) still misses every test — this lever speeds up *reruns*, not the first/cold suite. CI builds from a tagged seed (non-dirty stamp), so the per-child compiler-identity hash is skipped there (zero added cold cost).
- **Out of scope (per issue):** variant 2b (cache the pass/fail result), CI `build/cache/test-bin` save/restore plumbing, `--shard I/N`, and in-process `--jobs N` — all tracked separately in the proposal.
- A known small gap (documented): the key folds the compiler identity, not the runtime objects, so a `runtime/native/src/*.c` edit *without* a compiler rebuild could serve a stale binary. `make check`'s `--no-test-cache` is the backstop; in practice a runtime edit goes through `make compile`, which busts the identity.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XvVLXNzQ2Yt2itBjj4vG8m)_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,14 @@ endif
 export SAILFIN_AGENT_REPORT
 AGENT_REPORT := bash scripts/agent_report.sh
 
+# Extra flags appended to every `$(NATIVE_BIN) test ...` invocation
+# (#1230). Empty by default so the local `make test` inner loop uses the
+# per-test binary cache and gets fast incremental reruns. The `make
+# check` full-suite gate overrides this to `--no-test-cache` so the
+# merge/seedcheck gate always cold-builds every test binary — the cache
+# can never mask a test-compile regression.
+TEST_BIN_CACHE_FLAGS ?=
+
 help:
 	@echo "Common Sailfin tasks"
 	@echo ""
@@ -193,9 +201,9 @@ test-impl:
 	if [ "$${SAILFIN_AGENT_REPORT:-}" = "1" ]; then \
 		mkdir -p build; \
 		set -o pipefail; \
-		$(NATIVE_BIN) test compiler/tests/unit compiler/tests/integration compiler/tests/e2e capsules --json | tee build/agent-test.test.jsonl || rc=$$?; \
+		$(NATIVE_BIN) test compiler/tests/unit compiler/tests/integration compiler/tests/e2e capsules $(TEST_BIN_CACHE_FLAGS) --json | tee build/agent-test.test.jsonl || rc=$$?; \
 	else \
-		$(NATIVE_BIN) test compiler/tests/unit compiler/tests/integration compiler/tests/e2e capsules || rc=$$?; \
+		$(NATIVE_BIN) test compiler/tests/unit compiler/tests/integration compiler/tests/e2e capsules $(TEST_BIN_CACHE_FLAGS) || rc=$$?; \
 	fi; \
 	$(MAKE) test-e2e-sh || rc=$$?; \
 	exit $$rc
@@ -521,7 +529,7 @@ check-impl:
 		exit 1; \
 	fi
 	@echo "[check] running test suite on first-pass binary (early gate)..."
-	@$(MAKE) test NATIVE_BIN=build/native/sailfin
+	@$(MAKE) test NATIVE_BIN=build/native/sailfin TEST_BIN_CACHE_FLAGS=--no-test-cache
 	@echo "[check] first-pass tests passed — proceeding to seedcheck build..."
 	@echo "[check] verifying seed selfhost (stage2)..."
 	@# Stage2 routes through `sfn build -p compiler --work-dir <DIR>`
@@ -564,7 +572,7 @@ check-impl:
 	fi; \
 	echo "[check] seedcheck binary runs hello-world.sfn OK"
 	@echo "[check] running test suite with seedcheck binary (no fallbacks)..."
-	@$(MAKE) test NATIVE_BIN=build/native/sailfin-seedcheck
+	@$(MAKE) test NATIVE_BIN=build/native/sailfin-seedcheck TEST_BIN_CACHE_FLAGS=--no-test-cache
 	@echo ""
 	@echo "[check] stage2 tests passed — building stage3 for fixed-point comparison..."
 	@# Same shape as stage2 — driver `--work-dir` with
--- a/compiler/src/build_cache.sfn
+++ b/compiler/src/build_cache.sfn
@@ -805,13 +805,23 @@ fn test_bin_cache_root() -> string ![io] {
 }
 
 // Derive the content-addressed cache key for a single test's linked
-// binary. The dep set is the EXACT set of inputs the link consumes —
-// the resolver-produced dependency `.ll` files (`dep_ll_paths`, hashed
-// off disk) plus the staged `.sfn-asm` import contexts
-// (`native_texts`, hashed in memory). Deriving the key from the same
-// resolver output the link already uses is the correctness hinge: the
-// key can never miss a dep, because a changed dep changes a hash here
-// exactly as it changes the bytes the linker reads.
+// binary. The dep set is the test-source-specific set of inputs the
+// link consumes — the resolver-produced dependency `.ll` files
+// (`dep_ll_paths`, hashed off disk) plus the staged `.sfn-asm` import
+// contexts (`native_texts`, hashed in memory). Deriving the key from
+// the same resolver output the link already uses is the correctness
+// hinge for the test's OWN closure: a changed test source or dep
+// changes a hash here exactly as it changes the bytes the linker reads.
+//
+// Known gap: the link ALSO pulls in the assembled runtime capsule
+// objects + link libs (`_clang_link_test_cmd_with_deps`), which are NOT
+// folded into this key. Those are covered indirectly by
+// `compiler_identity` (a rebuilt compiler — which is how a runtime
+// source edit normally reaches a test binary, since the runtime is
+// rebuilt as part of `make compile` — busts every entry) and directly
+// by `--no-test-cache` on the `make check` / full-suite gate. A runtime
+// edit that is NOT accompanied by a compiler rebuild can therefore
+// serve a stale binary; the gate's cold build is the backstop.
 //
 //   key = sha256(
 //       sha256(test_source_bytes)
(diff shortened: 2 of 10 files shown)


mcereal:
Thanks @Copilot — all four addressed in `80ba04e`:

1. **`_sha256_of_string` tmp-path collision** — the `unique_suffix` now includes the per-invocation/per-child `scratch_root` (the multi-file parent mints a `mktemp -d` and hands each child its own `sub-N`; seedcheck pins `SAILFIN_TEST_SCRATCH`), so concurrent invocations and CI shards no longer share the deterministic `/tmp/.sfn_build_cache_hash_*` path. The suffix feeds only the temp filename, not the digest, so cache hits are unaffected.
2. **`chmod +x` failure** — now checked; a non-zero `chmod` leaves `cache_hit` false so the run falls through to a cold lower+link (counted as a miss) instead of exec'ing a non-executable file.
3. **Spec §11 over-claim** + 4. **status.md "exact link inputs"** — both qualified: the key does **not** fold the assembled runtime capsule objects / link libs. They ride on the `compiler_identity` component (a runtime edit normally reaches a test binary via `make compile`, which busts every entry), with `--no-test-cache` on the `make check` gate as the cold-build backstop. Same caveat added to `test_bin_cache_key`'s doc comment.

Verified: `make compile` self-hosts, e2e 4/4, `build_cache` unit 63/63, `fmt --check` clean.

---

Issue #1209, opened 2026-06-09
Epic: ownership/aliasing enforcement for native-runtime memory safety


mcereal:
## Epic

Implements the recommendation of [`docs/proposals/borrow-checking-1.0.md`](../blob/main/docs/proposals/borrow-checking-1.0.md) (#1207): a **bounded ownership/aliasing analysis enforced on the native runtime** as a 1.0 memory-safety floor — **not** a fourth pillar. The three pillars (effects, capabilities, concurrency) are unchanged. Motivated by #1205 (the grow-at-tip in-place-aliasing corruption reproduced in **both** the C runtime and its Sailfin port).

## ✅ Decisions locked (2026-06-09, maintainer)

The proposal's nine decision points (§8) are resolved — all on the recommended path:

| # | Decision | Resolution |
|---|---|---|
| **D1** | Adopt the stance reversal? | **Yes** — enforce ownership on the native runtime for 1.0 as a soundness floor (not a pillar). **Expansion mandate:** implement Option C as the *first rung of a deliberately extensible model*, not a ceiling — Sailfin will grow its ownership/borrowing story over time (a modern, agentic-development-era exploration of memory safety). "Sailfin doesn't have to be Rust." |
| **D2** | Enforced subset | **Option C** — unique-ownership / no-aliased-mutation / no-use-after-free over an `OwnedBuf` type + affine values, runtime-scoped. |
| **D3** | Blocker vs quality bar | **1.0 blocker**, satisfied narrowly by **Phase R1** (memory/string core enforced + #1205 determinism regression green). |
| **D4** | Ordering vs #822 | **Fix the aliasing contract before #822** deletes the C bodies. |
| **D5** | `OwnedBuf` type | **Yes** — new owned-buffer/`Slice` type family (extensible: views, region-scoped buffers can grow from it). |
| **D6** | `unsafe` boundary | **First-class `unsafe { }` / `unsafe fn`** (the `unsafe { }` block already parses; give it aliasing-boundary semantics). |
| **D7** | Pass placement | **Standalone `ownership_checker.sfn`** after effect-check (mirrors the effect-checker scope walk). |
| **D8** | M4 coupling | **M4 #965 consumes this ownership substrate** for channel-send / spawn-capture safe sharing. |
| **D9** | Doc surface | Ship the user-facing description as a **`reference/preview/` chapter** at 1.0; promote to a numbered spec chapter when Phase U (user-facing enforcement) ships. |

[...]

Issue #1205, opened 2026-06-09
bug(runtime): C-runtime in-place string optimization aliases stored strings → recurring nondeterministic IR corruption (root of #740/#892)


mcereal:
## Goal

Eliminate the **systemic** root cause behind a long-recurring, nondeterministic IR corruption: an *in-place* string append/concat optimization (grow-at-tip) mutates an arena buffer that a still-live string aliases, stomping unrelated emitted IR text (a `declare` line, a `bitcast … to %Type` line) under heap-layout shifts. The classic signature is a `0x40`→`0xE7` byte clobber (`@`/`%` flips to a UTF-8 lead byte, rendering as a CJK glyph e.g. `缯`).

> **Status update (2026-06-20): the C runtime is now deleted, and this bug is NOT fixed by that deletion.** Issue #1205 always called this out — "this is NOT a 'the C runtime will be deleted, so it goes away' bug." That has now been confirmed against the tree: with `sailfin_runtime.c` (#1434) and `sailfin_arena.c` (#1432) gone, the unsound grow-at-tip optimization **lives on entirely in the Sailfin runtime**. This issue is therefore re-homed from the C-retirement epic (#1308) to the ownership/aliasing epic (**#1209**), where its structural fix (Phase R1 / E8) belongs. The body below has been updated to drop the now-deleted C file/line/env references.

## Why this is filed as a systemic issue

Fixed twice at the **call-site** layer, kept recurring in **new** paths:

- #740 → #741 (`36e5cdc`) made the common `declare`-collection case deterministic.
- #892 (closed 2026-06-03) patched the effects/`declare` substring-view path (`effects.sfn`).
- 2026-06-08/09: recurred in the struct-field **type-name** rendering in `core_member_lowering.sfn` (`bitcast %NativeModule* … to 缯*` in a self-host `program.ll`, `clang: expected type`). Nondeterministic — a clean rebuild was correct.

Whack-a-mole because the fixes patch individual aliasing call-sites while the **underlying unsound in-place optimization stays live**.

## Root cause (file-cited, current Sailfin-only locations)

The type name is computed correctly and once (`compiler/src/llvm/type_context.sfn`). The corruption is the in-place buffer mutation in the **Sailfin runtime** (the C copies that this issue originally cited are deleted):

[...]

PR #1279, opened 2026-06-11
feat(compiler): in-place-mutation + use-after-free rules (E0902/E0903/E0906)

### LINKED ISSUE #1215: feat(compiler): in-place-mutation + use-after-free rules (E0902/E0903)
## Goal
Complete the 1.0 ownership subset: prove unique ownership at in-place mutation and disposal. An in-place mutator (`OwnedBuf.grow/append/set`) or disposer (`free`/drop) requires the receiver to be `Owned` with no other live alias — closing the #1205 hazard class structurally.

## Scope
**In:**
- `E0902` — in-place mutation of a possibly-aliased buffer (the #1205 grow-at-tip case).
- `E0903` — use after free / after drop (e.g. the `rc.sfn` release-to-zero-then-use hazard).
- `E0906` — raw-pointer escape into safe code outside an `unsafe` region (the E2 boundary).
- Two-span diagnostics + fix-its per §3.3; enforcement scoped to owned/affine values.
- Regression: a program that mutates an aliased buffer in place is rejected; the unique-owner path compiles.

**Out:**
- Migrating `runtime/sfn/**` onto the rules (E8/E9 — the consumers).
- Affine wrapper single-use (E7).

## Acceptance Criteria
- [ ] In-place mutation of an aliased buffer emits `E0902`; use-after-free emits `E0903`; raw-pointer escape emits `E0906`.
- [ ] The unique-owner grow-at-tip path (an `OwnedBuf` with no live alias) compiles clean.
- [ ] Regression coverage for all three codes + negative cases.
- [ ] `ulimit -v 8388608 && make compile` self-hosts.
- [ ] `ulimit -v 8388608 && make test` passes.

## Files Affected
- `compiler/src/ownership_checker.sfn` — mutation/UAF/escape rules
- `compiler/src/diagnostics*.sfn` — `E0902`/`E0903`/`E0906`
- `compiler/tests/unit/` + `compiler/tests/integration/`

## Verification
```bash
ulimit -v 8388608 && make compile && make test
```

## Size
M

## Type
Feature

## Blocked by
#1214

## Required in pinned seed
None for this issue. **This issue is the seed gate for E8**: once it merges, a seed cut + `/pin-seed` is required before E8 (runtime migration under enforcement) can self-host. Hence `seed-blocker`.

## Context
E6 of epic #1209 (D2/D3 = Option C, Phase R1 blocker). Closes the #1205 hazard class structurally — proving uniqueness makes grow-at-tip sound. Proposal §3.2, §4.2.

mcereal:
## Summary
- Completes E6 of the ownership epic (#1209): adds `E0902` (in-place mutation of a possibly-aliased buffer), `E0903` (use-after-free / after-drop), and `E0906` (raw-pointer escape into safe code) to the standalone ownership checker, closing the #1205 hazard class structurally.
- Adds a `"freed"` lattice state and a `_scope_after_call` classifier (disposer → mutator-method → generic + extern-escape), and threads the program-level `extern fn` name set through `OwnershipScope` for the E0906 boundary check.
- All three rules are scoped to owned/affine bindings (`OwnedBuf`/`Affine`/`Linear`). The un-migrated runtime passes raw `*u8`/`Cast` args (never a bare owned identifier), so the rules stay dormant on it and the compiler self-hosts unaffected.

Closes #1215

## Rule semantics
| Code | Fires on | Stays clean on |
|---|---|---|
| `E0902` | `x.grow/.append/.set(...)` on a **moved/freed** owned receiver (stale handle) | in-place mutation through the **unique live owner** (sound grow-at-tip) |
| `E0903` | use of a binding after `free(x)`/`drop(x)`/`x.free()`/`x.drop()` | disposal with no subsequent use |
| `E0906` | a bare **owned** identifier passed into an `extern fn` **outside** `unsafe` | the same escape **inside** `unsafe { }`; raw `*u8`/`Cast` args |

Two-span diagnostics (offending use + invalidating move/free) follow the existing `_make_finding` shape; `E0906` is a single-span boundary check.

## Acceptance Criteria
- [x] In-place mutation of an aliased buffer emits `E0902`; use-after-free emits `E0903`; raw-pointer escape emits `E0906`.
- [x] The unique-owner grow-at-tip path (an `OwnedBuf` with no live alias) compiles clean.
- [x] Regression coverage for all three codes + negative cases (10 new unit tests, 3 integration tests).
- [x] `ulimit -v 8388608 && make compile` self-hosts.
- [x] `ulimit -v 8388608 && make test` passes (full suite green locally; CI re-validates).

## Verification
```bash
ulimit -v 8388608 && make compile          # self-host gate: PASS (exit 0)
build/native/sailfin test compiler/tests/unit/ownership_checker_test.sfn        # PASS (all 27 tests)
build/native/sailfin test compiler/tests/integration/ownership_e6_test.sfn      # PASS (all 3 tests)
build/native/sailfin fmt --check compiler/src/ownership_checker.sfn ...          # clean
```
Self-host fence explicitly spot-checked: `free(p)` over a raw `*u8` is untracked; `ownedbuf.sfn` reads `self` fields (no mutator/disposer method calls on owned receivers), so no rule fires on the runtime.

## Files Changed
- `compiler/src/ownership_checker.sfn` — `"freed"` state, `_scope_after_call` classifier, `_consume_call_args`, `_gather_extern_names`, `extern_names` threaded through `OwnershipScope`; extended `_make_finding`/`_check_read`/`_consume_identifier`.
- `compiler/tests/unit/ownership_checker_test.sfn` — 10 new tests (E0902/E0903/E0906 positives + negatives + the self-host fence).
- `compiler/tests/integration/ownership_e6_test.sfn` — new end-to-end acceptance coverage.
- `docs/status.md` — ownership rows updated to reflect E6 enforcement.

## Notes
- E6 is the **seed gate for E8** (runtime migration under enforcement): once this merges, a seed cut + `/pin-seed` is required before E8 can self-host (hence `seed-blocker`).
- Design produced by the compiler-architect; reviewed by the code-reviewer (approved). Stale-receiver model chosen for E0902 (Slice-view alias tracking is deferred to Phase U per `ownedbuf.sfn`'s documented seam).

---
_Generated by [Claude Code](https://claude.ai/code/session_012WFCwdo6oFEELW2eCNDc1q)_


mcereal:
Re: the E0906 / `_gather_extern_names` scope (Copilot review comments on `ownership_checker.sfn:949` and `status.md:131)` — good catch on the doc overstatement; fixed in 5587562.

On the enforcement scope itself, this is deliberate for E6 rather than an oversight, and I've documented it precisely:

- **The doc row was overstated** — corrected to say the boundary is enforced for **locally-declared** externs, with the prelude/implicit-extern gap called out as a known, deliberate false negative.
- **Why not load prelude externs now:** `prelude_globals.load_prelude_global_names` returns *all* implicitly-available callables (fns, `let`s, externs) and is `[io]` (it FS-scans `runtime/sfn/**`). Wiring it into the ownership pass would (a) be semantically wrong — E0906 would then fire on a bare owned arg to any prelude *function* (e.g. `owned_buf_len(buf)`), not just externs; and (b) force this currently-pure analysis to take an IO dependency and invert module layering. Closing the gap properly needs an *extern-only* prelude name source, which is a follow-up.
- **The common bypass is already blocked upstream:** passing a bare owned value to a `* u8`-typed extern parameter (`memcpy`/`malloc`) is a typecheck error *before* the ownership pass runs, so the realistic residual gap is an extern declared elsewhere with an owned-typed parameter — narrow.
- **Self-host-safe:** the narrow scope only ever yields false *negatives*, never false positives, so it can't break the un-migrated runtime's self-host.

Tracking the widening (implicitly-linked extern coverage) as a follow-up under epic #1209.

---

mcereal:
Re: the mutator-method model on `ownership_checker.sfn:910` (live-owner `.append` non-consuming) — this is deliberate, and I've added a comment (23bba85) reconciling it with the shipped surface. Reasoning:

The proposal carries **two distinct append surfaces**, and the checker supports both as complementary (not contradictory):

- **§4.2 `owned_buf_append(self: OwnedBuf, ...) -> OwnedBuf`** — the shipped E3 (#1264) consume-and-return free function. A call passes `buf` as a bare-identifier argument, which the **move rules already treat as a move** (E0901/E0904): the source binding dies and the returned buffer is the new unique owner. So this surface is fully guarded — by the move half, not E0902.
- **§3.2 `OwnedBuf.grow/.append/.set`** — framed verbatim as *in-place mutators* whose receiver "must be `Owned` with no other live alias." That's an in-place `&mut self` borrow that does **not** consume. E0902 guards it: a stale (moved/freed) receiver is the violation; the unique live owner grows in place soundly.

So `buf.append(x); return buf;` being legal is correct *for the in-place mutator surface* — and it's specifically what the issue's acceptance criterion "the unique-owner grow-at-tip path … compiles clean" requires. Making `.append` consuming would instead turn that clean path into a use-after-move and collapse E0902 into the move rules, removing its distinct positive case.

There is no `.append` *method* on `OwnedBuf` in tree today, so neither E0902 nor this model fires on current runtime code — it's dormant on the real surface and self-host-safe. **E8 (the runtime migration under enforcement) settles which surface the migrated runtime adopts**; the checker is ready for both. I've documented this distinction at the call site and in the file header.

---

Issue #1218, opened 2026-06-09
refactor(runtime): migrate rc/mem/array to OwnedBuf ownership


mcereal:
## Goal
Extend Phase R1 enforcement to the rest of the memory core: `runtime/sfn/memory/rc.sfn`, `runtime/sfn/memory/mem.sfn`, `runtime/sfn/array.sfn` — re-expressing their raw-pointer in-place and `free` paths over `OwnedBuf`/unique ownership (or explicit `unsafe`).

## Scope
**In:**
- `rc.sfn`: prove no live alias before the release-to-zero `free` (`E0903`).
- `mem.sfn`: `copy_bytes`/`bounds_check`/`get_field`/`free` annotated `unsafe` or re-expressed over `OwnedBuf`/`Slice`.
- `array.sfn`: in-place push/grow behind unique ownership; element storage `unsafe`-internal.

**Out:**
- Concurrency modules (E11).
- The arena/string core (E8 — already done).

## Acceptance Criteria
- [ ] `rc.sfn`/`mem.sfn`/`array.sfn` pass ownership enforcement (no `E0902`/`E0903`/`E0906`).
- [ ] Existing runtime tests for these modules still pass (rc roundtrip, arena, array push).
- [ ] `ulimit -v 8388608 && make compile && make check` clean.

## Files Affected
- `runtime/sfn/memory/rc.sfn`, `runtime/sfn/memory/mem.sfn`, `runtime/sfn/array.sfn`
- their `compiler/tests/e2e/` harnesses

## Verification
```bash
ulimit -v 8388608 && make compile && make check
```

## Size
M

## Type
Refactor

## Blocked by
#1217

## Required in pinned seed
#1215 (E6) — same seed gate as E8; the ownership pass must be seeded. If E8's seed cut already covers it, no additional cut needed.

## Context
E9 of epic #1209. Completes the memory-core migration after E8. Proposal §5.1.

PR #1307, opened 2026-06-12
fix(test): isolate signal-kill retries + surface lost crash diagnostics (#1303)

### LINKED ISSUE #1303: test(flaky): atomic_fence_test / atomic_load_store_test SIGSEGV (139) under concurrent make check
## Goal
Diagnose and fix the intermittent `SIGSEGV` (exit code 139) in two atomic-intrinsic unit tests when run under the full concurrent `make check` suite. They currently degrade `make check`'s first-pass unit run to `148/150` even though both pass deterministically in isolation.

## Symptoms
- `compiler/tests/unit/atomic_fence_test.sfn` — `exit code 139, retrying once`
- `compiler/tests/unit/atomic_load_store_test.sfn` — `exit code 139, retrying once`

Observed during the `make check` first-pass (first-pass binary) unit suite:
```
[test] RETRY: atomic_fence_test.sfn (exit code 139, retrying once)
[test] RETRY: atomic_load_store_test.sfn (exit code 139, retrying once)
...
═══ unit: 148/150 passed, 2 failed ═══
  atomic_fence_test.sfn
  atomic_load_store_test.sfn
```
The remaining suites pass clean (integration 34/34, e2e 5/5, capsules 35/35).

## Evidence it is flakiness, not a real regression
- Both tests pass **3/3 deterministically** when run directly on the same first-pass binary:
  ```
  atomic_fence_test       run1/2/3 exit=0
  atomic_load_store_test  run1/2/3 exit=0
  ```
- The test harness already carries a **built-in single-retry path** for exactly these two tests — i.e. this is a previously-observed flaky condition.
- 139 = `128 + SIGSEGV(11)`: a runtime crash of the compiled test program, surfacing only under the concurrent suite's memory pressure (the toolchain self-caps at 8 GiB `RLIMIT_AS` on Linux, inherited by the parallel per-module emit/link children).

## Hypotheses to investigate
1. Memory pressure under parallel `make check` (many concurrent `clang`/`llvm-link` children sharing the inherited 8 GiB cap) pushing one of these test compiles/links over the cap mid-run → crash rather than a clean OOM diagnostic.
2. A genuine but rare data race / ordering issue in the atomic-intrinsic lowering or its runtime support, only reproduced under load.
3. Harness-level: the retry counts the test as failed even after a passing retry (the summary says `2 failed` despite the retry) — the retry accounting may itself be wrong.

## Acceptance Criteria
- [ ] Root cause identified (memory-pressure crash vs. real race vs. harness accounting).
- [ ] `make check` first-pass unit suite is `150/150` reliably across repeated runs (no flaky 139).
- [ ] If the fix is harness-side retry accounting, a passing retry no longer counts as a suite failure.

## Files Affected
- `compiler/tests/unit/atomic_fence_test.sfn`
- `compiler/tests/unit/atomic_load_store_test.sfn`
- the test runner / retry harness (`scripts/` test driver, `Makefile` check target)
- possibly `compiler/src/llvm/` atomic-intrinsic lowering if a real race is found

## Verification
```bash
make check    # first-pass unit suite must be 150/150, no SIGSEGV(139) retries
```

## Size
S

## Type
Bug

## Context
Surfaced while landing #1218 (E9, ownership epic #1209). The two failures are unrelated to that PR (IR-identical `unsafe` wrapping of rc/mem/array; the atomic tests don't import those modules) — filed separately so #1218 isn't blocked on pre-existing flakiness.

mcereal:
## Summary
- **Defer signal-kill (137/139) retries to an isolated serial tail** in the pooled `sfn test` runner. The old inline retry re-spawned *while up to `jobs-1` other children were still live* — under the same memory pressure that crashed the first attempt — so it tended to crash again. That is the flaky `148/150` first-pass unit signature: the two heavy-lowering atomic tests sit at **adjacent sorted indices 10/11**, so the sliding window co-resides them and they hit the memory peak together. Replaying after the pool drains matches the isolation where they pass **3/3 deterministically**.
- **Surface lost crash diagnostics.** A SIGSEGV child never flushes its stdio buffer, so its captured stream re-emits empty — the failure was counted (non-zero exit) and the file named in the suite banner, but with no inline attribution. Emit a synthetic `[test] FAIL: <file> (exit code N — child terminated by signal; captured output lost…)` for any signal-killed child whose output was lost, across the **pooled**, **pooled-retries-disabled**, and **multi-file-serial** reap paths.
- **Null-check `sfn_alloc_struct`** (defense-in-depth): host OOM made the arena/`calloc` path return NULL, which the lowered call sites GEP+store into → bare SIGSEGV. Abort with a `[sailfin-oom]` marker instead (composes with #1218's `unsafe` raw region). Plus a pre-retry unlink of the reused scratch `_subframe_summary.json` so a double-crash can't read a stale partial in `--json`.

Closes #1303

## Root cause
Not a real race and not broken accounting (both refuted by inspection): a **memory-pressure crash** + **ineffective retry**. Test files are sorted alphabetically; of the 13 unit files that call the arena-heavy `lower_to_llvm_ir_from_text`, only `atomic_fence_test`/`atomic_load_store_test` are an adjacent pair, guaranteeing they share a pool window under `jobs>1`. The retry then re-ran under the same pressure. (Per-process `RLIMIT_AS` is not the binding constraint — host RAM across N children is.)

## Acceptance Criteria
- [x] Root cause identified — memory-pressure crash of co-scheduled heavy-lowering tests + a retry that ran under the same pressure (not a race, not harness accounting).
- [x] First-pass unit suite reaches 150/150 reliably — addressed by isolating the retry; transient 137/139 now self-heals on the drained-pool replay. *(Full-suite green confirms on Linux CI — see Verification note.)*
- [x] A passing retry no longer counts as a suite failure — the deferred tail reassigns `rc` and accounts the file exactly once from the post-retry outcome.

## Verification
```bash
make compile                          # self-host PASS (with all changes)
sfn check / sfn fmt --check           # clean on both files
build/native/sailfin test atomic_fence_test.sfn      # 3/3 isolation exit=0
build/native/sailfin test atomic_load_store_test.sfn # 3/3 isolation exit=0

# Synthetic deterministic-crash fixture (null-store SIGSEGV) — harness logic:
#   pooled (retries on):       deferred → tail retry → synthetic FAIL → exit 139, 1 failed (counted once)
#   pooled (SAILFIN_TEST_RETRY=0): main-loop FAIL → synthetic diag → exit 139
#   multi-file serial (--jobs 1):  serial FAIL → synthetic diag → exit 139
#   companion passing test:    unaffected (1/1)
```
**Note on full-suite validation:** this flake is a **Linux/CI phenomenon** — `RLIMIT_AS` is a no-op on macOS and the dev box has ample RAM, so the full `make check` 150/150 run validates on CI, not locally. The local box additionally hit a degraded CommandLineTools toolchain today (Homebrew `llvm@17` clang can't link the reinstalled SDK; built via Apple clang), which inflated the local full-suite past its timeout. The flake **was** reproduced locally (both atomic tests crash under full `--jobs` pressure, pass in isolation), and the harness paths are validated deterministically via the synthetic fixture above.

## Files Changed
- `compiler/src/cli_commands.sfn` — deferred-retry serial tail, `_is_signal_exit`/`_emit_crash_diagnostic` helpers wired into all three reap fail paths, stale-summary pre-unlink.
- `runtime/sfn/memory/mem.sfn` — `sfn_alloc_struct` OOM null-check + `[sailfin-oom]` abort.

## Deferred follow-up
Regression coverage for the retry-accounting path was considered but deferred: a deterministic crash-then-pass fixture needs fragile cross-run state and would inject a deliberately-crashing test into the suite. The accounting is verified by review + the synthetic fixture here.

### DIFF
--- a/compiler/src/cli_commands.sfn
+++ b/compiler/src/cli_commands.sfn
@@ -761,6 +761,27 @@ fn _should_retry_test(exit_code: int, retries_disabled: boolean) -> boolean {
     return false;
 }
 
+// #1303: a test child that dies by signal (SIGSEGV/SIGABRT/SIGBUS/
+// SIGKILL → exit codes > 128) never flushes its stdio buffer, so the
+// parent re-emits an EMPTY captured stream — the crash leaves no
+// `test:` banner, no assertion, no phase trace. The failing file is
+// still named in the suite banner and the run still exits non-zero (the
+// signal exit code is caught by `handle_wait`), but a reader sees an
+// unexplained gap rather than an attributed failure. `timeout`'s own
+// 124/125 codes are NOT signal deaths (the child was killed by the
+// wrapper, which prints its own message) and stay below this threshold.
+fn _is_signal_exit(exit_code: int) -> boolean { return exit_code > 128; }
+
+// Emit a synthetic, attributed FAIL line for a signal-killed child whose
+// own output was lost to the unflushed buffer. Gated by the caller on
+// `!json_output` (the jsonl stream stays pure; the `--json` summary
+// already counts the file via the `sf_counts` missing-summary fallback).
+fn _emit_crash_diagnostic(basename: string, exit_code: int) -> void ![io] {
+    print("[test] FAIL: " + basename + " (exit code " + number_to_string(exit_code)
+        + " — child terminated by signal; captured output lost to the "
+        + "unflushed crash buffer)");
+}
+
 // Resolve the path to this compiler binary so the multi-file runner
 // can re-invoke itself (`<self> test <file>`) once per test for
 // process isolation. Mirrors `_resolve_self_path` in `cli_main.sfn`:
@@ -1588,6 +1609,14 @@ fn handle_test_command(args: string[], runtime_root: string, binary_dir: string)
             // file and eat the parallelism win.
             let pool_base_env = _pool_env_snapshot();
             let mut handles: int[] = [];
+            // #1303: indices (and their original exit codes) of children
+            // that hit a signal-kill (137/139) retry. Their retry is
+            // deferred to a serial tail after the pool drains so it runs
+            // in isolation rather than under the same memory pressure
+            // that crashed the first attempt — see the deferral comment
+            // in the reap block below.
+            let mut retry_pending: int[] = [];
+            let mut retry_rc: int[] = [];
             let mut launched: int = 0;
             let mut ti: int = 0;
             loop {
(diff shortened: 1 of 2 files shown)


mcereal:
Addressed both review comments in e7d2b0c9:

1. **`_is_signal_exit` range** — now bounded to `(128, 192]` (128 + max POSIX signal 64) instead of `> 128`, so a normal process exiting 193..255 isn't misattributed as a signal crash. 137/139 stay covered.
2. **Diagnostic wording** — no longer asserts total output loss; reworded to "output may be truncated (crash before stdio flush)" since a child may flush some lines before the signal.

Self-host re-verified and the synthetic deterministic-crash fixture still surfaces the diagnostic, exits 139, and counts the file exactly once.

Issue #1363, opened 2026-06-15
track: M1.A.2 — flip the `string` LLVM type-mapping from `i8*` to the `{i8*, i64}` SfnString aggregate


mcereal:
## Goal

Give "M1.A.2" — the `string` `i8*` → `{i8*, i64}` (SfnString aggregate) type-mapping flip — a real tracking home. Until now it has lived only as the comment tag **"M1.A.2"** scattered across ~27 files with no issue, repeatedly cited as a blocker but never trackable. Filed per the explicit ask in #1283 ("M1.A.2 still deserves its own tracking issue when that workstream is opened").

## Background

M1 (#450) locked the string **literal / aggregate shape** (M1.A) but deliberately carved out the **type-mapping flip** (M1.A.2). Today the language-level `string` type still lowers to a bare pointer, not the by-value aggregate:

- `compiler/src/llvm/type_mapping.sfn:513` — `if normalized == "string" { return "i8*"; }` (verified 2026-06-15). The very next lines (`:514–516`) carry a comment block describing the `{i8*, i64}` aggregate as the *pending* M1.A.2 end-state, with a matching guard in `map_return_type`.
- `compiler/src/cli_main.sfn:123` — M1.A.2 reference in the driver.
- The runtime carries a **by-value `SfnString {i8*, i64}` C ABI** already (e.g. `string.concat`/registry rows), but the Sailfin frontend cannot take/return an `SfnString` aggregate by value, so every runtime accessor body in `runtime/sfn/string.sfn` is forced to trampoline through `* u8` and recover the length by NUL-scan instead of reading `s.len` off the aggregate.

## Why this matters (what it unblocks)

1. **#1283 Gap 2 — sound `slice -> Slice`.** A non-owning `Slice { data: text.data + start, len: end - start }` is unrepresentable until `string` is the `{i8*, i64}` aggregate (and the immediate-codepoint pseudo-pointer encoding is retired in #822). M1.A.2 is one of its two named dependencies.
2. **The runtime string family ports (epic #1308 — #1315, #1318).** Their architect-spec bodies (`(SfnString, i64) -> i64` direct `s.data[idx]` indexing, `s.len` field reads, `memchr` over `s.data[0..s.len]`) all assume the aggregate. Today they keep the `* u8` + NUL-scan proof-of-life shape because M1.A.2 hasn't landed.
[...]

Issue #1315, opened 2026-06-12
refactor(runtime): port string byte/codepoint/grapheme accessors to Sailfin


mcereal:
## Goal
Port the string accessor family to real UTF-8 walks in Sailfin and make the C bare emission-target defs `static` so emission binds to the Sailfin bodies (the pointer-read half of R1).

## ⚠️ Re-scoped 2026-06-13 (#1309-class audit) — link-ownership flip + Files-Affected gap
- Two symbol classes here:
  - **Emission targets** (`native_signature`): bare `sfn_str_codepoint`/`sfn_str_grapheme_at`/`sfn_str_grapheme_count` (+ `sfn_str_byte_at`/`sfn_str_find_byte`) are **C-defined, not in the header** (`inHdr=0`), no internal C callers → define real Sailfin bodies (rename the `sfn_str_sfn_*` infix to bare or add bare wrappers) + `static`-ify the C defs. `.c`-only, no header edit.
  - **Legacy `sailfin_runtime_*`** (`char_code`, `grapheme_count`, `string_length`) are consumed only via `string.sfn`'s façade externs — making the bodies real **removes those externs**. These C symbols **stay** (e.g. `sailfin_runtime_string_length` has **9 internal C callers**) until #822. Do **not** static-ify them.
- **Files-Affected gap:** add `sailfin_runtime.c` (static-ify the bare `sfn_str_*` targets). **No layout hazard** — `SfnString` is by-value `{i8*, i64}`; verify the port emits the identical aggregate.

## Scope
**In:**
- `runtime/sfn/string.sfn` — real `byte_at`/`find_byte`/`codepoint`/`grapheme_at`/`grapheme_count` bodies (UTF-8 walks; `find_byte` via `memchr`), bare-named; drop the `sailfin_runtime_{char_code,grapheme_*,string_length}` façade externs.
- `runtime/native/src/sailfin_runtime.c` — `static`-ify the bare `sfn_str_{byte_at,find_byte,codepoint,grapheme_at,grapheme_count}` defs.

**Out:**
- Allocating ops (concat/slice/from_*/to_*) → #1318. Legacy `sailfin_runtime_*` C bodies (retired in #822).

## Acceptance Criteria
- [ ] `nm`: `string.sfn` defines the bare accessor symbols; `sailfin_runtime.o` no longer defines them.
- [ ] `strings` tests pass; the lexer (hot consumer) still self-hosts.
[...]

Issue #1308, opened 2026-06-12
Epic: retire the C runtime — flip residual canonical sfn_* bodies to Sailfin, then delete sailfin_runtime.c


mcereal:
## Epic

Complete the C→Sailfin runtime migration and physically delete the C runtime. This epic **re-homes the C-deletion stories (#822, #823) out of the M4 concurrency epic (#965)** so #965 is purely about concurrency, and adds the residual-symbol-flip work that #822's concurrency-only acceptance criteria never captured.

### Why this epic exists (empirical evidence)

A maintainer asked "can we just delete the C runtime now?" The audit (`docs/runtime_audit.md`) and the registry suggested most symbols were ported, but an **object-level link experiment** proved otherwise. Zeroing `c-sources`/`ll-sources` in `runtime/native/capsule.toml` and relinking (`make clean-build && make compile`) fails with **49 undefined C-defined symbols** still demanded by the Sailfin-compiled objects:

```
nm -u build/sailfin/*.o  ∩  {symbols defined in sailfin_runtime.c + sailfin_arena.c}  =  49 symbols
```

Root cause: most "ports" under `runtime/sfn/**` are **proof-of-life façades that still `extern`-call C primitives** — e.g. `string.sfn`'s `sfn_str_sfn_from_codepoint` forwards to the C `sfn_str_from_codepoint`; `arena.sfn`'s wrappers call the C `sfn_arena_alloc`/`sfn_arena_global`; `filesystem.sfn` forwards to `sailfin_adapter_fs_*`. The registry's `native_signature` flips routed *emission* to canonical names, but those canonical bodies (and the primitives the Sailfin wrappers call) are **still C**. So the C runtime (9,228 + 357 lines) remains load-bearing far beyond concurrency.

Only **1 of the 49** (`sailfin_runtime_serve`) is concurrency/M4-scoped — already tracked under #965. The other **48 are M2/M3 residue** that no open issue covers.

### The 49 blocking symbols, by family

| # | Family | Symbols | Target Sailfin home | Size |
|---|---|---|---|---|
[...]

Issue #1323, opened 2026-06-13
feat(http): typed serve vertical slice — sfn_serve_framed + lowering gate + capsule serve


mcereal:
Wave 2 of epic #1321 — the first user-visible slice. Design: `docs/proposals/sfn-http-capsule.md` §2, §3.1–3.3, §5.

Deliberately bundled (runtime + compiler + capsule + example + e2e) because the compiler-lowering gate's single consumer is the capsule `serve`; splitting would add a PR with no standalone value. **No seed cut needed** — the compiler change and its consumer land in one `make compile` pass and the compiler never calls `serve` itself.

## Goal

Ship a working typed HTTP server: `serve(handler: fn(Request) -> Response, port: int = 8080) -> void [net, io]` whose handler's status codes and headers actually reach the wire.

## Scope

**In:**
- **Runtime (additive, Sailfin-only):** `sfn_serve_framed(handler: * u8, port: i32)` in `runtime/sfn/concurrency/serve.sfn` — same accept loop as `sfn_serve`, but the handler's return is the **full pre-framed HTTP response** sent verbatim (`null` → `500`). Reuse the `_serve_*` helpers. Migration-mechanism 3 (additive): no C definition, no header, no shared-layout hazard.
- **Compiler (gate, Wave 2 only):** in `compiler/src/llvm/expression_lowering/native/core.sfn` (~line 1415), gate the bespoke `serve(` lowering on `serve` **not** resolving to a user-defined/imported function (consult the `functions` map in scope; fall through to generic call lowering when found). Bare `serve(handle, 8080)` with no import stays unaffected.
- **Capsule:** replace the `serve` print-stub in `capsules/sfn/http/src/mod.sfn` with the typed signature; store the handler code pointer in a module-global i64 slot (the `future.sfn` `_sfn_g_*_addr` pattern); add `_sfn_http_trampoline(raw: string) -> string` calling `parse_request` → user handler (via `* fn` cast) → `serialize_response`; declare `extern fn sfn_serve_framed(...)`. Emit a best-effort diagnostic / doc note that the handler must be a **top-level function** (closures unsupported in v0).
- **Example:** rewrite `examples/web/http-server.sfn` to the return-based `fn(Request) -> Response` form.
[...]

PR #1369, opened 2026-06-16
WIP(M1.A.2): flip scalar string ABI to {i8*, i64} aggregate (#1363)

mcereal:
## Summary

Implements **M1.A.2** (#1363): flips the language-level `string` type from a bare `i8*` to the by-value `{i8*, i64}` (data, len) **SfnString aggregate**, so the runtime string family (#1308) can take/return strings by value with their length — unblocking #1315/#1318.

> **Status: WIP — self-hosts but the suite is not yet green.** One well-characterized blocker remains (cyclic layout fixed-point) plus ~20 stale-ABI test updates. Full writeup: `docs/proposals/m1a2-string-abi-status.md`.

## What works (verified)

- **Self-hosts cleanly.** A cold `make clean-build && make check` emits **0 invalid-IR / parallel-emit errors** — the compiler compiles itself with the new ABI.
- **Lands in ONE PR — no seed cut.** Disproves the "mandatory seed cut" verdict: pass-1 internals stay `i8*` (seed-built); the `exec.sfn` `* u8` self-host externs keep pass-2 extern/def shapes consistent.
- **Scalar `string` → `{i8*, i64}`** across all three type-mapping modules (`type_mapping`, `core_type_mapping`, `statement_type_mapping`) — fixes the param-`.length`-returns-0 bug (the call side previously passed only the data pointer, leaving the length register undefined).
- **Container scoping:** `string[]` elements, optional `string?`, module globals, and catch-bindings stay `i8*`; only scalar locals/params/returns flip. Adds the `{i8*,i64}`↔pointer cast arm and drops the let-binding `i8*` clamp.
- Struct string fields → 16-byte `{i8*, i64}`; canonical layout-table **leaf/struct sizes regenerated and fixed-point-verified**.

## Remaining blocker (needs a layout-engine change)

`canonical_type_layouts()` hand-pins ~28 by-value sizes that encoded the old `string=8` ABI. The **cyclic enum cluster** (`Statement` embeds `ForClause`/`MatchCase` by value) does not converge by hand because the table is consulted before the `visiting` cycle-guard — each manual iteration inflates the next (`Statement` 96→128→184…). **Disabling the table is worse** (157 vs ~125 regressions), so it's load-bearing.

**Recommended:** make the canonical table self-computing to a fixed point (iterate `calculate_record_layout` until stable, or derive the cyclic enum payloads from the LLVM type sizes LLVM already resolves) rather than hand-pinning.

Also remaining: ~20 tests that deliberately pin the **old** `i8*` ABI must be updated to the aggregate (e.g. `abi_value_return_test` asserts `fn s() -> string` lowers to `define i8* @s` *and not* `{i8*, i64}`; `test_struct_field_separator.sh` pins `%Pair = type { i8*, double }`).

## Files changed
- `compiler/src/llvm/type_mapping.sfn`, `.../core_type_mapping.sfn`, `.../statement_type_mapping.sfn` — the flip + container clamps
- `compiler/src/llvm/expression_lowering/native/core_literals_lowering.sfn` — cast arm
- `compiler/src/llvm/lowering/instructions_let.sfn` — drop the i8* clamp
- `compiler/src/emit_native_layout.sfn` — string field size + canonical table regen
- `runtime/sfn/platform/exec.sfn` — self-host externs as `* u8`
- `docs/proposals/m1a2-string-abi-status.md` — full status / continuation notes

Refs #1363

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ku1m92FXinfpusfmhjLz7N)_


mcereal:
### Status: linux-x86_64 fully green; one macOS-only follow-up tracked

The M1.A.2 scalar `string` → `{i8*, i64}` flip is complete and **self-hosts in one pass (no seed cut)**. Every regression it surfaced was a genuine `i8*`↔aggregate boundary-coercion gap, root-caused and fixed; only 3 changes were legitimate ABI-pin updates (`abi_value_return`, `test_drop_emission_try_catch`, `escape_promotion_test`).

The last and deepest one was the CI `e2e-sh-2` HTTP failure: the **runtime-object cache key wasn't ABI-aware**, so a seed-compiled `i8*` prelude was silently reused by the flipped compiler's `{i8*,i64}` call sites (`find_char` returned garbage → empty `req.path` → 404). Fixed by bumping the cache tag so runtime objects recompile once under the aggregate ABI.

**CI on `1b63102`:** linux-x86_64 is **fully green** (unit-a/b, int-e2e-caps, all e2e-sh including the HTTP shard). The macOS body-drain test also passes.

**One remaining red check — macOS-arm64 `e2e-sh-2`** — is the `#1323` *"imported serve lowers to the capsule fn"* gate: on Darwin the imported sfn/http `serve` isn't staged into the lowering `functions` list, so the builtin `@sfn_serve` hijacks the call (`core.sfn:1431`). It is **deterministic, macOS-only, passes on `main`, and orthogonal to the string flip** (an import-context staging defect). Filed as **#1370** with the root cause located; it needs a Darwin runner to verify a fix, so it is tracked separately rather than blocking this PR.

---

Issue #1855, opened 2026-07-03
bug(lowering): bare struct literal via type annotation lowers to zeroinitializer (fields dropped)


mcereal:
Discovered while fixing #1838 (union field-name global). Splitting out because it is an independent, broader lowering bug in a different semantic unit.

## Goal

A bare struct literal resolved through a type annotation (`let x: T = { field: val, ... }`) must populate its fields, not lower to a zero value.

## Root cause / symptom

The **bare** literal form assigned via a declared-type annotation lowers to `zeroinitializer` / `null` — the field values from the literal are never stored. The **explicit** form (`T { field: val }`) works correctly.

Minimal repros (all with the current compiler, post-#1838):

```sfn
// concrete struct — bare form, via annotation → prints "Person: " (empty)
struct Person { name: string; }
fn main() [io] {
    let p: Person = { name: "Bob" };
    print("Person: {{p.name}}");
}
```

```sfn
// same struct — explicit form → prints "Person: Bob" (correct)
let p = Person { name: "Bob" };
```

```sfn
// intersection type — the examples/advanced/unions.sfn `admin` case
interface Admin { isAdmin: boolean; }
interface User { name: string; }
type AdminUser = Admin & User;
fn main() [io] {
    let admin: AdminUser = { name: "Alice", isAdmin: true };
    print("Admin: {{admin.name}}");   // prints "Admin: " (empty)
}
```

Emitted IR for the concrete case shows the literal collapsing to a zero value:

```
%l0 = alloca %Person
store %Person zeroinitializer, %Person* %l0   ; fields from { name: "Bob" } dropped
```

So the annotation-directed bare literal is not routed through the field-population path the explicit `T { ... }` form uses.

## Scope

- In: lowering of a bare struct literal (`{ field: val, ... }`) whose target struct/interface/intersection type comes from a `let`/param type annotation — populate its fields the same way the explicit `T { ... }` form does.
- Out: the explicit `T { ... }` form (already correct); dynamic member-access field-name global emission (#1838, fixed); the `x is T` type-guard operator (#1753).

## Acceptance criteria

- [ ] `let p: Person = { name: "Bob" }; print("{{p.name}}")` prints `Bob`.
- [ ] `examples/advanced/unions.sfn` prints `Admin: Alice` (currently `Admin: `).
- [ ] The intersection-type case (`type AdminUser = Admin & User`) populates fields.
- [ ] Regression test in `compiler/tests/`.
[...]

PR #1916, opened 2026-07-04
fix(typecheck): complete E0828 object-literal coverage via a unified TypeckCtx walk context

### LINKED ISSUE #1900: fix(typecheck): normalize array/generic-head object-literal targets + cover parameter defaults (E0828)
## Goal
Close the two contained slices of SFEP-0039 §3.2's residual `E0828` surface: (1) normalize **array** (`Named[]`) and **generic-instantiation-head** (`Iface<T>`) object-literal targets so the classifier resolves them correctly, and (2) walk **parameter-default** initializers so `fn f(p: SomeInterface = { ... })` is rejected. Both live where the classifier already has `top_level` in scope, so no context-threading is needed here.

## Context
Follow-up to #1860 / PR #1899, which shipped `E0828` at the primary `let` site and left the normalization + parameter-default gaps flagged in the `_object_literal_head_type` doc comment. Originally scoped as all residual positions; the return-position (#1904) and lambda-body/return (#1905) slices were split out because they require wide/deep context-threading. This issue is the contained remainder.

`check_object_literal_target` and `_object_literal_head_type` in `typecheck_types.sfn` are the extension points; `_object_literal_head_type` today strips only a trailing `?`.

### Design decision — array targets use `E0828`, not a new code (Rust/Go-consistent)
An object literal against an array target (`let x: Named[] = { ... }`) is rejected with **`E0828`**, reusing the existing rule with a dedicated array message branch — **not** a new diagnostic code. Rationale: Rust (`E0308`) and Go (assignability) both reject this as a compile-time mismatch through their *general* machinery rather than minting a bespoke per-shape code; E0828's rule is already "object literal without a concrete-struct target," and an array is one more non-struct shape. The interface / un-inferable / array distinction is a *message* branch (already the `is_interface` bool), not a code distinction.

⚠️ **Correctness subtlety — array vs generic head differ:**
- **Array `T[]`**: fire `E0828` **regardless of whether `T` is a struct**. `let x: Named[] = {}` is wrong *even when `Named` is a struct*, because `{}` never yields an array. Detect a trailing `[]` and **short-circuit to the array branch before the head-symbol lookup** — do NOT strip `[]` and classify the element.
- **Generic head `Base<...>`**: strip the `<...>` block, then classify the **base** name normally. `Box<int>` → `Box` is a struct → **OK (no diagnostic)**; `Iface<int>` → `Iface` is an interface → `E0828`. (`find_symbol` currently matches the raw `Iface<T>` text and misses — base-name normalization is the fix.)

## Scope
**In:**
- Extending the object-literal target classifier (`_object_literal_head_type` / `check_object_literal_target`) to normalize array and generic-instantiation-head target forms, with the array short-circuit semantics above.
- An array-specific `E0828` message branch (names the array type; steers to building elements and collecting into an array literal `[...]`).
- Walking `Parameter.default_value` object-literal initializers in the function-signature check and emitting `E0828` there via the existing classifier.
- Regression coverage per new target form and for the parameter-default position.

**Out:**
- Return-position object literals (#1904) and lambda-body / lambda-return object literals (#1905) — both need context-threading and are separate.
- Any change to `E0829` or the shipped `let`-site `E0828` / optional-`?` handling (#1899).
- Reference sigils (`&Named` / `&mut Named`) — not Sailfin syntax, out of scope.
- Structural / dynamic object-model machinery (permanently out per SFEP-0039).

## Acceptance Criteria
- [ ] `let x: Named[] = { ... }` reports `E0828` (array target) **whether or not `Named` is a struct**.
- [ ] `let x: Iface<T> = { ... }` reports `E0828` when `Iface` is an interface; `let x: Box<int> = { ... }` still compiles when `Box` is a struct.
- [ ] `fn f(p: SomeInterface = { ... })` reports `E0828`; a concrete-struct-typed default still compiles.
- [ ] The shipped `let`-site behavior (interface / un-inferable / optional-`?` / concrete-struct) is unchanged.
- [ ] `assert_does_not_compile` regression tests for the array target, generic-interface-head target, and parameter-default position added under `compiler/tests/`.
- [ ] SFEP-0039 §5 self-host audit: no `compiler/src` idiom is newly rejected (pre-audit found none).
- [ ] `make compile` passes.
- [ ] `make test` passes.
- [ ] `sfn fmt --check` clean on every touched `.sfn`.

## Files Affected (advisory map — non-binding, expected to drift)
- compiler/src/typecheck_types.sfn — `_object_literal_head_type` (array `[]` short-circuit + generic-head base extraction, reusing the `<...>` depth scan already in `extract_generic_argument_block` / `split_top_level_ampersand`), `make_object_literal_requires_struct_diagnostic` (array message branch), and `check_function_signature` (walk `parameter.default_value`). `top_level` is already a parameter of `check_function_signature`.
- compiler/tests/ — regression coverage.

## Verification
```bash
make compile && make test
```

## Size
S

## Type
Bug

## Required in pinned seed
None

## Blocked by
None

## Design
SFEP-0039 (docs/proposals/0039-nominal-object-model.md) §3.2 + §3.6 (E-code allocation). The array-target-uses-E0828 decision is recorded above per the "nominal, Rust/Go-shaped object model" framing of the SFEP.

## Context
Original issue scope was decomposed at grooming into this slice + #1904 (return position) + #1905 (lambda). Recommended pickup order: this issue first (its normalization lands in the shared classifier that #1904/#1905 reuse), then #1904 and #1905 in parallel. Flagged by the code review on PR #1899.

### LINKED ISSUE #1904: fix(typecheck): emit E0828 on return-position object literals
## Goal
Emit `E0828` when a bare object literal `{ ... }` appears in a `return` statement whose enclosing function declares a non-`struct` return type (interface / array / generic-interface head), closing that slice of the residual silent-miscompile surface SFEP-0039 §3.2 targets.

## Context
Follow-up to #1860 / PR #1899, which shipped `E0828` at the primary `let` site only. Split out of #1900 (which was too large for one session — three positions with very different threading cost). The classifier `check_object_literal_target` in `typecheck_types.sfn` already exists and is reused unchanged; the work here is **threading the enclosing return type into the statement walk** so `return { ... }` can be classified.

Today no return-type context reaches the expression walk. `check_try_operator`'s `enclosing_return_type` parameter is **dead** (no live caller) — it is evidence the live walk has no return-type context, not machinery to reuse. `check_function_body` calls `check_block` without passing `signature.return_type`, and the `ReturnStatement` branch of `check_statement` only recurses into `walk_expression`.

## Scope
**In:**
- Threading the enclosing function's declared return type into the block/statement walk so a `return` expression can see it.
- Emitting `E0828` (via the existing `check_object_literal_target`) when a `return`ed expression is an object literal and the enclosing return type resolves to a non-struct target (interface / un-inferable). Array and generic-interface-head return targets are covered once #1900's normalization lands in the shared classifier; do not duplicate that normalization here.
- Regression coverage for the return position (interface return target rejected; concrete-struct return target still compiles).

**Out:**
- Any change to the `let`-site `E0828`, the target-form normalization, or parameter-default coverage (all in #1900).
- Lambda-body / lambda-return object literals (separate issue).
- `E0829` (already complete at all value positions).
- Any expression-type inferencer / `check_try_operator` live-wiring beyond what a return-type context requires (#829 is a separate workstream).

## Acceptance Criteria
- [ ] `fn f() -> SomeInterface { return { ... }; }` reports `E0828` (interface return target).
- [ ] A function returning a concrete `struct` via a `return { ... }` still compiles (the sanctioned #1855 path).
- [ ] A top-level statement context (no enclosing function / return type) does not spuriously fire and does not crash — the threaded return type defaults to absent there.
- [ ] `assert_does_not_compile` regression test(s) for the return position added under `compiler/tests/`.
- [ ] SFEP-0039 §5 self-host audit: no `compiler/src` return-position idiom is newly rejected (pre-audit found none).
- [ ] `make compile` passes.
- [ ] `make test` passes.
- [ ] `sfn fmt --check` clean on every touched `.sfn`.

## Files Affected (advisory map — non-binding, expected to drift)
- compiler/src/typecheck.sfn — the block/statement walk (`check_function_body`, `check_block`, `check_statement` and its `ReturnStatement` branch; every nested-block-recursing variant — `If`/`For`/`Match`/`With` — is a call site the threaded return-type context must pass through) and the top-level scope entry where no enclosing return type exists.
- compiler/src/typecheck_types.sfn — `check_object_literal_target` (reused as-is; no change expected).
- compiler/tests/ — regression coverage.

## Verification
```bash
make compile && make test
```

## Size
M

## Type
Bug

## Blocked by
None

## Required in pinned seed
None

## Design
SFEP-0039 (docs/proposals/0039-nominal-object-model.md) §3.2. Reuses the `E0828` diagnostic and `check_object_literal_target` shipped in #1899.

## Context
Split from #1900. Recommended pickup order: after #1900 (so the shared classifier already normalizes array/generic-head targets), but not hard-blocked — the helper exists today.

### LINKED ISSUE #1905: fix(typecheck): emit E0828 in lambda-body / lambda-return object literals
## Goal
Emit `E0828` for bare object literals in lambda bodies (`let` inside a lambda) and lambda return-type positions, so the lambda-body twin of the primary `let` site gets the same non-struct-target rejection, closing that slice of SFEP-0039 §3.2's residual silent-miscompile surface.

## Context
Follow-up to #1860 / PR #1899. Split out of #1900. The lambda-body expression walk (`walk_statement_expressions` / `walk_block_expressions`, reached from `walk_expression`'s `Lambda` branch) already calls the intersection and bare-fn checks but **not** `check_object_literal_target` — because it lacks the `top_level: SymbolEntry[]` symbol table the classifier needs. The lambda `Lambda` branch already has `expression.return_type` in hand, so a lambda return-type check is locally plausible once `top_level` is threaded down. This is a wide-but-mechanical parameter-threading change.

## Scope
**In:**
- Threading `top_level: SymbolEntry[]` down into the expression walk so the lambda-body `let` site and the lambda return-type site can call the existing `check_object_literal_target`.
- Emitting `E0828` at the lambda-body `let` position and the lambda return-type position when the target resolves to a non-struct (interface / un-inferable; array + generic-head covered by #1900's normalization in the shared classifier).
- Regression coverage for both lambda positions.

**Out:**
- The `let`-site, return-position, normalization, and parameter-default work (in #1900 and the return-position issue).
- `E0829` (already wired at the lambda-body `let` and lambda-return positions; unaffected).
- Any semantic change to lambda typing beyond the object-literal target check.

## Acceptance Criteria
- [ ] A `let x: SomeInterface = { ... }` inside a lambda body reports `E0828`.
- [ ] A lambda declared to return an interface, whose body returns `{ ... }`, reports `E0828`.
- [ ] The same positions targeting a concrete `struct` still compile.
- [ ] `assert_does_not_compile` regression test(s) for the lambda positions added under `compiler/tests/`.
- [ ] SFEP-0039 §5 self-host audit: no `compiler/src` lambda-position idiom is newly rejected (pre-audit found none).
- [ ] `make compile` passes.
- [ ] `make test` passes.
- [ ] `sfn fmt --check` clean on every touched `.sfn`.

## Files Affected (advisory map — non-binding, expected to drift)
- compiler/src/typecheck.sfn — the expression walk that must carry `top_level`: `walk_expression` and its recursive call sites, plus `walk_block_expressions` / `walk_statement_expressions` (the lambda-body `let` twin) and the `Lambda` branch's return-type site.
- compiler/src/typecheck_types.sfn — `check_object_literal_target` (reused as-is; no change expected).
- compiler/tests/ — regression coverage.

## Verification
```bash
make compile && make test
```

## Size
M

## Type
Bug

## Blocked by
None

## Required in pinned seed
None

## Design
SFEP-0039 (docs/proposals/0039-nominal-object-model.md) §3.2. Reuses the `E0828` diagnostic and `check_object_literal_target` shipped in #1899.

## Context
Split from #1900. Recommended pickup order: after #1900 (shared classifier normalization), but not hard-blocked. The threading is mechanical (identical extra argument at each recursive `walk_expression` call site) — a good `implementer` slice once specced.

mcereal:
## Summary

Closes the residual `E0828` ("bare object literal `{ ... }` requires a concrete-`struct` target", SFEP-0039 §3.2) surface. `E0828` previously fired only at the primary `let` site (#1899); a bare object literal targeting an interface/array/generic-head — or returned from a function or lambda — slipped through silently.

Rather than land the three follow-up issues as three separate ad-hoc parameter-threadings, this PR introduces **one unified `TypeckCtx` expected-type + typing-environment context** threaded through both typecheck walk families — the same abstraction Rust (`FnCtxt` + `Expectation`) and Go (assignment-target propagation) use. Every value position now reads the expected type from the *same* channel, so the residual-position backlog is retired structurally instead of extended one issue at a time. Design record: **SFEP-0041** (`docs/proposals/0041-typeck-expected-type-context.md`, Implemented).

> Note: #1900 was found closed-as-completed but never actually implemented (only a docs PR had merged); it was reopened and its work is delivered here.

## What changed

Staged so each step self-hosts on its own:

- **Stage A — classifier (`typecheck_types.sfn`), closes #1900.** Array targets (`Named[]`) fire `E0828` **regardless of element type** (an object literal never yields an array), via a short-circuit before the head-symbol lookup; generic-instantiation heads (`Iface<...>`) are classified by base name; parameter-default initializers (`fn f(p: Iface = { ... })`) are walked in `seed_parameter_scope`. The diagnostic maker gained a dedicated array message branch.
- **Stage B — statement walk (`typecheck.sfn`), closes #1904.** New `TypeckCtx` record + `make_typeck_ctx`/`ctx_with_bindings`/`ctx_enter_body` helpers. `check_statement`/`check_block`/`check_function_body` migrated onto it via destructure-at-entry (bodies stay behavior-identical). A `return { ... }` is classified against the enclosing function's declared return type (`ctx.enclosing_return_type`), gated on `!is_top_level`.
- **Stage C — expression walk (`typecheck.sfn`), closes #1905.** The expression family (`walk_expression`/`walk_match_pattern`/`walk_block_expressions`/`walk_statement_expressions`) migrated onto `TypeckCtx`. The `Lambda` branch enters the body via `ctx_enter_body`, installing the lambda's return type; lambda-body `let` and lambda-`return` object literals are now classified (per-lambda scoping preserved for nested lambdas). Unifying the record is what finally gives the lambda-body walk `top_level` + the enclosing return type — the structural fix #1905 needed.

The single classifier `check_object_literal_target` is unchanged and remains the only decision point; the named-`struct` construction path (`Struct` AST variant, the #1855 path) stays exempt throughout.

## Validation

- `make compile` (self-host) green after **each** stage.
- `sfn fmt --check` clean on every touched `.sfn`; `sfn check` clean.
- `compiler/tests/unit/typecheck_object_model_test.sfn`: **34/34** (let / array / generic-head / parameter-default / return / lambda-body / lambda-return, with struct-target negative controls and a nested-lambda per-scope test); `effect_raw_failclosed_test.sfn` adapted for the new signature (**11/11**). Walk-sensitive suites (scope/shadowing, expression walk, conditionals, `check_try_operator`, lambda capture, effect checker, concurrency) all green.
- SFEP-0039 §5 self-host audit: `make compile` confirms no `compiler/src` idiom (return-position or lambda-body) is newly rejected — the compiler contains no bare `return { ... }` object literals.

## Docs

- `docs/proposals/0041-typeck-expected-type-context.md` — new SFEP (Implemented) + registry row.
- `docs/status.md` and SFEP-0039 §3.2 — record complete `E0828` coverage across value positions.

Closes #1900
Closes #1904
Closes #1905

---
_Generated by [Claude Code](https://claude.ai/code/session_017qdF3dXtNJExYNSUkmkR2E)_

### DIFF
--- a/compiler/src/typecheck.sfn
+++ b/compiler/src/typecheck.sfn
@@ -419,12 +419,17 @@ fn register_top_level_symbol(statement: Statement, existing: SymbolEntry[]) -> S
 // reaches a nested `return` (the #1904 fix). `expected_type` is the
 // general local channel reserved for future positions (call args, field
 // defaults) so they plug in without another signature change.
+// `in_function_body` marks that the walk is inside a fn/lambda body, so a
+// `return { ... }` is classified there but a top-level `return` — or one
+// in a top-level nested block (`if`/`with`/`{}`), which is `!is_top_level`
+// yet has no enclosing function — is exempt.
 struct TypeckCtx {
     bindings: SymbolEntry[];
     imports: ImportSymbolTable;
     interfaces: Statement[];
     top_level: SymbolEntry[];
     is_top_level: boolean;
+    in_function_body: boolean;
     scope_start: int;
     enclosing_return_type: TypeAnnotation?;
     expected_type: TypeAnnotation?;
@@ -438,6 +443,7 @@ fn make_typeck_ctx(bindings: SymbolEntry[], imports: ImportSymbolTable, interfac
         interfaces: interfaces,
         top_level: top_level,
         is_top_level: is_top_level,
+        in_function_body: false,
         scope_start: 0,
         enclosing_return_type: null,
         expected_type: null
@@ -445,7 +451,10 @@ fn make_typeck_ctx(bindings: SymbolEntry[], imports: ImportSymbolTable, interfac
 }
 
 // Enter a nested block scope: swap in the block's bindings + shadowing
-// boundary, drop to non-top-level, preserve the ambient return type.
+// boundary, drop to non-top-level, preserve the ambient return type and
+// the in-function-body flag. `expected_type` is a parent-imposed *local*
+// expectation, so it is cleared at a statement/block boundary — it must
+// not leak into unrelated statements once the channel becomes active.
 fn ctx_with_bindings(ctx: TypeckCtx, bindings: SymbolEntry[], scope_start: int) -> TypeckCtx {
     return TypeckCtx {
         bindings: bindings,
@@ -453,15 +462,16 @@ fn ctx_with_bindings(ctx: TypeckCtx, bindings: SymbolEntry[], scope_start: int) 
         interfaces: ctx.interfaces,
         top_level: ctx.top_level,
         is_top_level: false,
+        in_function_body: ctx.in_function_body,
         scope_start: scope_start,
         enclosing_return_type: ctx.enclosing_return_type,
-        expected_type: ctx.expected_type
+        expected_type: null
     };
 }
 
 // Enter a fn/lambda body: install its parameter-scope bindings and the
-// declared return type as the ambient `enclosing_return_type`, clearing
-// any local expected type.
+// declared return type as the ambient `enclosing_return_type`, mark the
+// in-function-body context, and clear any local expected type.
 fn ctx_enter_body(ctx: TypeckCtx, bindings: SymbolEntry[], return_type: TypeAnnotation?) -> TypeckCtx {
     return TypeckCtx {
         bindings: bindings,
@@ -469,6 +479,7 @@ fn ctx_enter_body(ctx: TypeckCtx, bindings: SymbolEntry[], return_type: TypeAnno
         interfaces: ctx.interfaces,
         top_level: ctx.top_level,
         is_top_level: false,
+        in_function_body: true,
         scope_start: 0,
         enclosing_return_type: return_type,
         expected_type: null
@@ -513,6 +524,7 @@ fn check_statement(statement: Statement, ctx: TypeckCtx) -> ScopeResult {
     let imports = ctx.imports;
     let top_level = ctx.top_level;
     let is_top_level = ctx.is_top_level;
+    let in_function_body = ctx.in_function_body;
     let enclosing_return_type = ctx.enclosing_return_type;
     if statement.variant == "VariableDeclaration" {
         let registration = register_local_symbol(bindings, statement.name, "variable", statement.span, scope_start);
@@ -792,10 +804,12 @@ fn check_statement(statement: Statement, ctx: TypeckCtx) -> ScopeResult {
     if statement.variant == "ReturnStatement" {
         let return_diags = walk_expression(statement.expression, ctx);
         // A `return { ... }` is classified against the enclosing function's
-        // declared return type (#1904). Gate on `!is_top_level`: a top-level
-        // `return` has no enclosing function, so `enclosing_return_type` is
-        // absent and the un-inferable branch must not fire spuriously.
-        if is_top_level {
+        // declared return type (#1904). Gate on `in_function_body`, not
+        // `!is_top_level`: a `return` in a top-level nested block (`if`/`with`/
+        // `{}`) is `!is_top_level` yet has no enclosing function, so its
+        // `enclosing_return_type` is absent and the un-inferable branch must
+        // not fire there.
+        if !in_function_body {
             return ScopeResult {
                 bindings: bindings,
                 diagnostics: return_diags
--- a/compiler/src/typecheck.sfn
+++ b/compiler/src/typecheck.sfn
@@ -516,7 +516,7 @@ fn check_statement(statement: Statement, ctx: TypeckCtx) -> ScopeResult {
     let enclosing_return_type = ctx.enclosing_return_type;
     if statement.variant == "VariableDeclaration" {
         let registration = register_local_symbol(bindings, statement.name, "variable", statement.span, scope_start);
-        let walk_diags = walk_expression(statement.initializer, bindings, imports);
+        let walk_diags = walk_expression(statement.initializer, ctx);
         let annotation_diags = check_bare_fn_type_annotation(statement.type_annotation, statement.name, statement.span);
         let intersection_diags = check_intersection_value_type_annotation(statement.type_annotation, statement.name, statement.span);
         let object_literal_diags = check_object_literal_target(statement.initializer, statement.type_annotation, statement.name, statement.span, top_level);
@@ -706,8 +706,8 @@ fn check_statement(statement: Statement, ctx: TypeckCtx) -> ScopeResult {
         };
     }
     if statement.variant == "ForStatement" {
-        let target_diags = walk_expression(statement.clause.target, bindings, imports);
-        let iter_diags = walk_expression(statement.clause.iterable, bindings, imports);
+        let target_diags = walk_expression(statement.clause.target, ctx);
+        let iter_diags = walk_expression(statement.clause.iterable, ctx);
         let body_diags = check_block(statement.body, ctx_with_bindings(ctx, bindings, scope_start));
         return ScopeResult {
             bindings: bindings,
@@ -715,13 +715,13 @@ fn check_statement(statement: Statement, ctx: TypeckCtx) -> ScopeResult {
         };
     }
     if statement.variant == "MatchStatement" {
-        let mut diagnostics = walk_expression(statement.expression, bindings, imports);
+        let mut diagnostics = walk_expression(statement.expression, ctx);
         let mut index: int = 0;
         loop {
             if index >= statement.cases.length { break; }
             let case = statement.cases[index];
-            diagnostics = diagnostics.concat(walk_match_pattern(case.pattern, bindings, imports));
-            diagnostics = diagnostics.concat(walk_expression(case.guard, bindings, imports));
+            diagnostics = diagnostics.concat(walk_match_pattern(case.pattern, ctx));
+            diagnostics = diagnostics.concat(walk_expression(case.guard, ctx));
             let arm_bindings = register_match_pattern_bindings(case.pattern, bindings);
             let _blk_diags = check_block(case.body, ctx_with_bindings(ctx, arm_bindings, scope_start));
             let mut _ci: int = 0;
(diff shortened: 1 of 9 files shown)


mcereal:
Thanks for the review — addressed in `6dbb024`.

**Fixed:**
- **Return gate precision** (`typecheck.sfn:799`): replaced the `!is_top_level` gate with a dedicated `in_function_body` flag on `TypeckCtx` (set by `ctx_enter_body`, preserved by `ctx_with_bindings`). A `return` in a top-level nested block is now correctly exempt rather than firing the un-inferable branch. (In practice a top-level `if`/`with` is itself a parse error, but the gate is now semantically precise regardless.)
- **`expected_type` scoping** (`ctx_with_bindings`, `:458`): now cleared at the block boundary instead of carried through, so a future parent-imposed expectation can't leak into unrelated statements once the channel is active.
- **Array detection** (`typecheck_types.sfn:428`): tightened to match the `[]` suffix specifically rather than any trailing `]`.
- **Generic-head message** (`typecheck_types.sfn:472`): now names the full head (`Container<int>`) verbatim while still resolving the symbol by base name.
- **Docs drift** (SFEP-0041 & SFEP-0039): corrected the `TypeckCtx` file location (it lives in `typecheck.sfn`, not `typecheck_types.sfn`, because `ImportSymbolTable` is imported there) and replaced the `typecheck.sfn:450` line references with symbol-based ones (`check_statement`'s `VariableDeclaration` branch).

Also added a regression test for a no-declared-return-type function returning a bare object literal (the `in_function_body` / un-inferable path).

**Deferred as out of scope — filed #1917:** the note that lambda parameters aren't in the lambda-body walk scope (E0420 false-positive when a param is called as a function). This is **pre-existing** — the `Lambda` branch has always walked the body with the outer bindings, and this PR preserves that exactly; it isn't introduced or regressed here. Repro confirmed (`let apply = fn(f, x) -> int { return f(x); }` → `E0420`), tracked in #1917 with a scoped fix.

`make compile` self-hosts and the full unit suite (196/196) is green with these changes.

---

Issue #1935, opened 2026-07-05
feat(compiler): nested / local function name as a first-class value (SFEP-0042 Sub-issue B)


mcereal:
## Design: SFEP-0042

`docs/proposals/0042-nested-function-declarations.md` (Accepted), §S4 + Appendix. **Sub-issue B** of what is now the **First-Class Function Values** project (formerly GitHub epic SailfinIO/sailfin#1609, which closed `not_planned` on 2026-07-07 as part of the Linear migration — *not* abandoned).

## Goal

Allow a nested/local function's **bare name** to be used in **value position** (assignment, passed as an argument, stored in a field) — lowering to the closure pair via the dispatch seam, and narrowing the `E0808`/[#1147](<https://github.com/SailfinIO/sailfin/issues/1147>) value-position guard for this case. Sub-issue A ([#1922](<https://github.com/SailfinIO/sailfin/issues/1922>), the by-name construct) landed 2026-07-06.

## Scope

**In:**

* Resolve a nested fn name in value position to its lifted top-level symbol and lower to `{fnptr, null-env}` (same path as a top-level named fn value).
* **Narrow** (never delete) the `#1147` `E0808` guard for the nested-fn value case once the lowering exists — SFEP-0030 D2 is explicit that the guard stays load-bearing for generic-fn-as-value, `& fn`, and non-`* u8` casts.

**Out:**

* The by-name call construct (Sub-issue A / [#1922](<https://github.com/SailfinIO/sailfin/issues/1922>)).
* Capturing nested fns — rejected by design. The accepted rule is the `E0421` capture rejection in SFEP-0042 §3.3; §6 is the alternatives section (it rejects Go-style capturing sugar and the let-lambda desugar), not the normative statement.

## Acceptance Criteria

- [ ] A nested fn name passed to a `fn(...)`-typed parameter / assigned to a `fn(...)` `let` compiles and dispatches correctly.
- [ ] Regression + e2e coverage.
- [ ] `make compile` and `make test` pass; `sfn fmt --check` clean.

## Ground truth (verified 2026-07-31)

Nothing of this issue is implemented; nothing of it is obsolete. Confirmed against the working tree:

[...]

Issue #1609, opened 2026-06-24
Epic: First-class function values (named fns, fn-typed struct fields, capturing closures as general params)


mcereal:
## Goal

Make function values genuinely first-class in user code: a value of type `fn(...) -> ...` can be **named-function-sourced**, **stored in a struct field**, **passed as an ordinary parameter**, and **invoked** — including when it is a **capturing closure**, in arbitrary user code (not just the runtime array-HOF descriptor seam).

This is a design/tracking **epic** — it needs an architect pass before fan-out into XS/S/M sub-issues.

## Why this is filed (the gap, ground-truthed)

The pieces shipped so far are narrow and don't compose into a usable first-class-function surface:

- **#1142 / #1146** shipped only the *referencing* half: `<fn> as *u8` lowers a named fn to a bare C-ABI code pointer.
- **#1147** shipped the **E0808 guard** that (correctly) *errors* on any un-lowerable function-in-value-position — i.e. it closes the silent-null hole but does **not** enable the feature.
- **#1118** (closed) shipped a **narrow** runtime-callable closure-apply seam — but only for `fn(...)`-typed **runtime array descriptor** params (`runtime_array_{map,filter,reduce}_fn`), pointer-width `i8*(i8*,i8*)` ABI only. General user code does not go through that seam.

Verified against `sfn 0.7.0-alpha.49+dev.6e0534b` (post-#1507):

| Form | Probe | Result |
|---|---|---|
| Named fn passed as a typed `fn(...)` value/param | `apply(dbl, 5)` | ❌ typecheck error (E0808-family guard, #1147) — referencing exists only as `<fn> as *u8` |
| `fn(...)`-typed **struct field** declaration | `struct Ops { compile: fn(int)->int }` | ⚠️ field *type* accepted by `check`; **populating** it with a named fn hits the named-fn wall; field-dispatch unverified |
| **Capturing** closure as a **general** `fn(...)` param (non-descriptor user code) | `apply(fn (x) { x + base }, 5)` | ❌ **miscompile** — LLVM `store i8* %t0, i8** %t5` → `'%t0' i64 but expected ptr` (filed separately, see sub-issues) |
| Array `map`/`filter`/`reduce` capturing closures (pointer-width int) | in-tree e2e | ✅ works (the #1118 seam) |
[...]

Issue #1389, opened 2026-06-17
fix(check): sfn check is green while sfn build fails at link


mcereal:
Part of #1385.

## Goal
Stop `sfn check` reporting success on programs that fail at `sfn build` link time (false-green inner loop).

## Symptom
The #1386 repro (`let mut xs: int[] = [];`) returns `checked 1 files: ok` from `sfn check`, then fails at `sfn build` link. Static analysis does not model module-global init emission.

## Scope
**In:** Decide and implement the right contract — either (a) `check` detects the unbuildable-global class, or (b) `check`'s guarantee is explicitly documented as "parse + typecheck + effect-check, not codegen/link", with the gap noted.
**Out:** Full codegen-in-check (out of scope; `check` stays fast).

## Acceptance Criteria
- [ ] After #1386 lands, the original false-green disappears naturally; add a test asserting `check` and `build` agree on a representative module-global program.
- [ ] If any check/build divergence remains, it is documented in the validation-ladder section of CLAUDE.md / docs.
- [ ] `make compile` / `make test` pass.

## Files Affected
- `compiler/src/` (check pipeline) and/or `docs/` — depending on chosen contract.
- `compiler/tests/`

## Size
- [x] **S**

## Type
type:bug / area:dx

PR #1939, opened 2026-07-05
feat(compiler): nested / local function declarations (SFEP-0042)

### LINKED ISSUE #1922: feat(compiler): nested / local function declarations called by name (SFEP-0042 Sub-issue A)
## Design: SFEP-0042

`docs/proposals/0042-nested-function-declarations.md` (Accepted). This is **Sub-issue A** of epic #1609 (first-class function values). Read the SFEP as the brief — do not re-derive the design.

> **History:** this issue was originally filed as a typecheck walk-scope bug on the hypothesis that statement-position `fn` produced a `Statement.FunctionDeclaration` that fell through `walk_statement_expressions`. Ground-truthing showed the parser does **not** produce that node at all — nested `fn` hits the `E0818` "unstructured expression" backstop. The real work is a small feature (parser + typecheck + emit + effect fix), re-scoped here under SFEP-0042.

## Goal

Support a non-capturing, block-hoisted, statement-position `fn name(...) -> T [effects] { ... }` **called by name** inside another function or lambda body. Clears the `E0818` backstop; delivers the whole user-visible feature except using a nested fn's bare name as a first-class value (that is Sub-issue B, gated on #1609).

## Scope

**In:**
- **S1 Parser** — new `fn`-name dispatch arm in `parse_block_statement` (`compiler/src/parser/statements.sfn`), two-token lookahead (`fn` + identifier) to disambiguate from anonymous lambda `fn(`; delegate to `parse_function`; wrap in `Statement.FunctionDeclaration`. Reject `async`/decorators at nested position cleanly; accept `unsafe fn`.
- **S2 Typecheck** — two-pass **block-hoist** pre-pass (register all sibling `FunctionDeclaration` names before walking bodies) in `check_block` + `walk_block_expressions`; new `FunctionDeclaration` arm in `walk_statement_expressions` that walks the body in a **restricted scope** (own params + module items + hoisted siblings, NOT enclosing locals) and emits new **E0421** on an enclosing-local reference (capture rejection, Rust-E0434 analog) with a `let f = fn(...) => ...` fix-it.
- **S3 Emit** — reuse the `lift_non_capturing_lambdas` shape (`compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn`) but lift to a **plain** top-level fn (no `__env`, no fat pointer); scope-qualified mangling `sfn_nested_<enclosing_chain>_<name>_<N>`; rewrite by-name call sites.
- **Effect fix** — correct `effect_checker.sfn:605` `FunctionDeclaration` arm that inlines a nested body's effects into the enclosing block; a declared-but-uncalled nested `fn foo() [io]` must impose no effect on its parent; a **call** requires transitive effects. Register nested-fn effect signatures for call-site checking.

**Out:**
- Nested fn as a first-class **value** (`{fnptr, null-env}`) — **Sub-issue B**, blocked by #1609.
- Capturing nested fns — rejected by design (use a `let`-lambda); SFEP-0042 §6.
- `async` / decorators on nested fns — deferred (rejected cleanly in v0).

## Acceptance Criteria
- [ ] Statement-position `fn helper() -> int { return 1; }` inside a function/lambda body parses (no `E0818`) and, when called by name, compiles and runs.
- [ ] Two sibling nested fns mutually recurse; a nested fn self-recurses (block-hoisting).
- [ ] A nested fn body referencing an enclosing `let` or parameter emits **E0421** with the fix-it; referencing its own param / module-level fn / sibling nested fn does not.
- [ ] Two same-name nested fns in one block emit the duplicate-symbol diagnostic (E0001).
- [ ] A declared-but-uncalled nested `fn foo() [io]` does not force `[io]` on its parent; a parent that calls `foo()` without `[io]` gets the effect diagnostic.
- [ ] Two different outer fns each declaring `fn helper()` both compile (distinct mangled symbols).
- [ ] Regression tests under `compiler/tests/{unit,integration,e2e}/`.
- [ ] `make compile` and `make test` pass; `sfn fmt --check` clean. Graduate SFEP-0042 to `Implemented`.

## Files Affected (advisory)
- `compiler/src/parser/statements.sfn` (S1)
- `compiler/src/typecheck.sfn`, `compiler/src/typecheck_types.sfn` (S2, E0421)
- `compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn` (S3)
- `compiler/src/effect_checker.sfn` (effect fix)
- `compiler/tests/{unit,integration,e2e}/`

## Required in pinned seed
_None — additive parser syntax bundled with its own consumers; `make compile` builds the new compiler from the old seed and it compiles the feature in the same self-host pass (SFEP-0042 §5, seed-dependency default-to-bundle)._

## Verification
```bash
make compile && make test
```

## Size
M

## Type
Feature

mcereal:
## Summary
- Adds statement-position `fn name(...) -> T [effects] { ... }` inside a function or lambda body — **non-capturing static items** (Rust model), block-hoisted for self/mutual recursion, lifted to plain top-level functions and called by name via static dispatch.
- Closes the pre-existing `E0818` "unstructured expression" wall for nested `fn` (the original #1922 premise — a typecheck walk-scope gap — turned out to be a parser gap; see the issue thread).
- Design record: **SFEP-0042** (`docs/proposals/0042-nested-function-declarations.md`, Accepted). This is **Sub-issue A** of epic #1609 (first-class function values); using a nested fn's bare name as a first-class *value* is Sub-issue B (#1935, gated on #1609).

Closes #1922

## Design decisions
- **Capture model — Rust, non-capturing.** A nested `fn` may reference its own params, module-level items, and sibling nested fns — never an enclosing local/parameter. Capturing one is rejected with the new **`E0421`** (Rust-`E0434` analog) and a fix-it pointing at `let f = fn(...) => ...` (Sailfin's existing capturing form). A same-named nested-fn parameter legally shadows an enclosing local.
- **Visibility — block-hoisted.** Sibling nested fns mutually recurse and a nested fn self-recurses regardless of declaration order (two-pass block walk), in contrast to the sequential `let` registration.

## Pipeline
- **Parser** (`parser/statements.sfn`): `fn`-name arm in `parse_block_statement` (two-token lookahead disambiguates from anonymous lambda `fn(`); `unsafe fn` handled; `async`/decorated nested fns fall through to a clean parse rejection.
- **Typecheck** (`typecheck.sfn`, `typecheck_types.sfn`): block-hoist pre-pass; non-capturing body scope with enclosing locals tagged `outer_local`; `E0421` capture rejection emitted from the shared `walk_expression` resolution choke point; lambda-body walk arm.
- **Emit** (`llvm/expression_lowering/native/lambda_lowering.sfn`): reuses the lambda-lift pre-pass to hoist nested fns to plain top-level functions with collision-safe `sfn_nested_<name>_<N>` mangling and block-scoped call-site rewriting; descends into struct method bodies.
- **Effects** (`effect_checker.sfn`): a nested declaration no longer inlines its effects into the parent; nested signatures are registered in a **routine-local** effect table (scope-correct, no global collision) for call transitivity; each nested body is effect-checked against its own row, including nested fns inside lambda bodies.

## Code review follow-ups (addressed in this PR)
The Opus code-review gate caught and this PR fixes:
- **Effect soundness (blocking):** nested-fn effect signatures were briefly registered into the global table (last-write-wins), which could strip a same-named top-level function's effects from unrelated callers. Now scope-local (deep-copied to avoid in-place mutation of the shared table).
- **Method-body nested fns:** were a green-`check`/broken-build (#1389 class) — now lifted correctly.
- **Lambda-body nested fns:** now effect-checked and registered (collector descends into lambda bodies).

## Acceptance Criteria
- [x] `fn helper() { ... }` in a function/lambda body parses (no `E0818`), compiles, and runs when called by name.
- [x] Sibling nested fns mutually recurse; a nested fn self-recurses.
- [x] Enclosing-local reference → `E0421` with fix-it; a same-named param shadows (no false positive).
- [x] Two same-name nested fns in one block → `E0001`.
- [x] A declared-but-uncalled nested `fn foo() [io]` imposes no effect on its parent; a *call* requires it transitively; a nested body using an undeclared effect is flagged.
- [x] Two different outer fns each declaring `fn helper()` both compile (distinct mangled symbols).
- [x] Regression tests under `compiler/tests/{unit,e2e}/`.
- [x] `make compile` self-hosts; `sfn fmt --check` clean.

## Map reconciliation
The issue's advisory `## Files Affected` named only `typecheck.sfn` (its premise was a typecheck-only walk-scope bug). Ground-truthing showed the real work spans **parser → typecheck → emit → effect-checker** (the parser never produced the `FunctionDeclaration` node the issue assumed). This is the true semantic scope of the Goal, re-scoped under SFEP-0042 with the issue re-groomed; not silent scope growth (documented on the issue thread and in the SFEP).

## Verification
```
make compile                     # self-hosts (from the pinned seed)
sfn fmt --check <touched files>  # clean
make test                        # unit 199/199, integration 42/42, capsules 38/38,
                                 # e2e 193/194 — the one non-pass is work_dir_parity_test.sfn
                                 # killed by the serial (--jobs 1) run's outer timeout;
                                 # it PASSES in isolation (2/2). Not a regression.
```
Nested-fn suites (all green): `nested_function_declaration_test.sfn` (unit, 11) · `nested_function_effect_test.sfn` (unit, 7, incl. the review collision/cross-contamination/lambda-body cases) · `nested_function_declaration_test.sfn` (e2e, 4, incl. struct-method-body build+run).

## Files Changed
- `compiler/src/parser/statements.sfn`
- `compiler/src/typecheck.sfn`, `compiler/src/typecheck_types.sfn`
- `compiler/src/llvm/expression_lowering/native/lambda_lowering.sfn`
- `compiler/src/effect_checker.sfn`
- `compiler/tests/unit/nested_function_declaration_test.sfn`, `compiler/tests/unit/nested_function_effect_test.sfn`, `compiler/tests/e2e/nested_function_declaration_test.sfn`
- `docs/proposals/0042-nested-function-declarations.md`, `docs/proposals/README.md`, `docs/status.md`

---
_Generated by [Claude Code](https://claude.ai/code/session_01KPqZAAsxw5JdFG6WNKaFUc)_


mcereal:
Thanks — addressed the Copilot review in `5de89ab`. Summary of dispositions:

**Fixed (same-name shadowing between nested fns):**
- **`_build_nested_fn_scope` dropping top-level-shadowed siblings** — now adds sibling/enclosing nested-fn names regardless of a same-named module symbol, ordered to shadow it. A nested `fn` shadowing a top-level function now resolves to (and self-recurses through) the nested one, consistent with the emit rename. New e2e test `shadows a same-named top-level function` (returns 2, not the top-level 1).
- **`_lookup_nested_rename` scanning start-first** — now scans innermost-first, so an inner nested `fn` shadows a same-named outer one at inner call sites. New e2e test `inner-scope nested fn shadows a same-named outer one`.

**Documented as v0 limitations + tracked in #1940** (both need scope-path-aware keying, not a bare-name key — a follow-up refactor, and both are narrow):
- **A `let`/parameter shadowing a same-named nested `fn`** (emit rewrites the call to the lifted symbol even though typecheck resolves it to the value binding).
- **Same-named nested fns in *sibling* blocks with differing effects** (routine-local effect table flattens by bare name, last-write-wins). Captured in SFEP-0042 §3.4.

**Docs status (SFEP `Accepted` vs `status.md` "Shipped"):** intentional, not an oversight. SFEP-0042 has two sub-issues; this PR is **Sub-issue A (call-by-name)**, which ships and self-hosts — hence `status.md`'s "Shipped (**call-by-name**)". **Sub-issue B** (nested fn name as a first-class value, #1935) is gated on epic #1609's `{fnptr, null-env}` work, so the SFEP stays `Accepted` until it lands (per the multi-issue-slice convention). Left as-is deliberately.

Distinct-name nested fns — the common case — are unaffected by the documented gaps. `make compile` self-hosts; all nested-fn suites pass (unit 11 + effect 7 + e2e 6).

---

PR #2061, opened 2026-07-09
feat(semver): version parse, §11 precedence ordering, floor check (SFN-169)

mcereal:
## Summary

Adds `compiler/src/semver.sfn` — the one genuinely new primitive the native
toolchain-pin feature (SFEP-0046 §3.2) needs. The tree could only *extract*
version strings (`version.sfn`); nothing compared or ordered them. This module
parses a semantic version, orders two versions per semver §11 precedence, and
answers a `>=` floor check. It has no dependency on the manifest/gate work and
unblocks both later phases (SFN-167 version check, SFN-168 dispatch guard).

All functions are `[pure]`; the module depends only on the prelude to keep the
self-host build graph minimal.

Fixes SFN-169

Design: SFEP-0046 (`docs/proposals/0046-toolchain-pinning.md`, Accepted) §3.2

## Changes

- **runtime/compiler primitive** — new `compiler/src/semver.sfn` exporting:
  - `struct SemVer { major, minor, patch: int; prerelease: string }` (build
    metadata after `+` parsed and ignored for ordering, semver §10)
  - `semver_parse(text) -> SemVer` — tolerant of a leading `v`; rejects
    malformed input, including leading-zero core fields per semver §2. A
    rejected parse returns a `major == -1` sentinel detectable via
    `semver_is_valid` (the return type is `SemVer`, not `Result`).
  - `semver_compare(a, b) -> int` — `-1/0/+1`, semver §11 precedence
  - `semver_satisfies_floor(running, pin) -> boolean` — accepts `>=`, rejects `<`
  - NPM-style range expressions intentionally omitted (SFEP-0046 §6); the schema
    leaves room for a later range predicate over `SemVer` with no schema change.
- **tests** — `compiler/tests/unit/semver_parse_test.sfn` (14 tests) and
  `compiler/tests/unit/semver_compare_test.sfn` (10 tests).

## Validation

- [x] `sfn fmt --check` clean on every touched `.sfn` file
- [x] `sfn check` (fast static analysis) passes — `checked 3 files: ok`
- [x] `make compile` — compiler self-hosts (`status: pass`)
- [ ] `make check` — not run; purely additive, non-structural module not yet
      imported by any consumer, so it cannot perturb existing behavior. The
      issue's verification (`make compile` + the unit tests) passes.
- [x] Regression coverage added under `compiler/tests/unit/` (24 tests, all pass)

## Notes for reviewers

- Reviewed by the Opus `code-reviewer`: §11 ordering and parse robustness
  verified correct (no panic/out-of-range paths — `substring` clamps, every
  index is bounds-guarded), effects sound, self-host low-risk. Two low-severity
  notes were addressed in-diff: core fields now reject leading zeros (strict
  semver §2), and `_parse_int`'s assumed-bounded input is documented.
- **Stage1 readiness intentionally omitted:** this is an internal toolchain
  primitive with no user-facing syntax or effect-enforcement surface. The
  end-to-end toolchain-pin feature (and its SFEP graduation) ships via SFN-167 /
  SFN-168, which wire this primitive into the manifest gate and dispatch guard.

## Map reconciliation

None — the advisory `## Files affected` map (`compiler/src/semver.sfn` +
`semver_parse_test.sfn` + `semver_compare_test.sfn`) matched the tree exactly.
One in-scope design decision: added `semver_is_valid` as the reject-detection
predicate, required to make the documented `major == -1` sentinel usable given
the non-`Result` return type.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CY8DUG5qifQrddeArzg68w)_

### DIFF
--- a/compiler/src/semver.sfn
+++ b/compiler/src/semver.sfn
@@ -1,0 +1,232 @@
+// compiler/src/semver.sfn
+// Semantic-version parsing and precedence ordering for the native toolchain.
+// The tree extracts version strings (version.sfn) but has no way to *order*
+// two versions or compare prereleases; the toolchain-pin feature (SFEP-0046)
+// needs exactly that. This module is the isolated, pure primitive: parse a
+// version, order two per semver §11, and answer a `>=` floor check. Build
+// metadata after `+` is parsed and ignored for ordering (semver §10).
+//
+// A malformed parse is reported by an invalid sentinel (`major == -1`, an
+// impossible value for a real version) via `semver_is_valid`, since the
+// return type is `SemVer`, not a `Result`. Range expressions (npm-style `^`
+// / `~`) are intentionally absent (SFEP-0046 §6); a range predicate can be
+// added later as a new function over `SemVer` without changing this schema.
+struct SemVer {
+    major: int;
+    minor: int;
+    patch: int;
+    prerelease: string;
+}
+
+fn _invalid() -> SemVer ![pure] {
+    return SemVer {
+        major: -1,
+        minor: -1,
+        patch: -1,
+        prerelease: ""
+    };
+}
+
+// True unless `v` came from a rejected/malformed parse.
+fn semver_is_valid(v: SemVer) -> boolean ![pure] { return v.major >= 0; }
+
+fn _is_digit(ch: string) -> boolean ![pure] {
+    if ch.length != 1 { return false; }
+    let c = char_code(ch);
+    return c >= 48 && c <= 57;
+}
+
+fn _is_ident_char(ch: string) -> boolean ![pure] {
+    if ch.length != 1 { return false; }
+    let c = char_code(ch);
+    if c >= 48 && c <= 57 { return true; }
+    if c >= 65 && c <= 90 { return true; }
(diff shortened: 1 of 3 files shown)


mcereal:
Thanks — all four Copilot review points were valid and are addressed in 9121876:

- **§9 leading zeros in numeric prerelease ids:** `_valid_prerelease` now rejects them (`1.0.0-01` no longer parses).
- **Overflow in numeric prerelease comparison:** `_compare_identifier` now orders numeric identifiers by length-then-lexical (no `_parse_int`), so arbitrarily long ids can't overflow — leading-zero rejection makes a longer digit run always the larger number.
- **Overflow in core fields:** `_valid_core_number` caps core fields at 18 digits (< 10^18, always fits i64), rejecting oversized manifest/tag input instead of wrapping.
- **Invalid-sentinel floor guard:** `semver_satisfies_floor` now returns false unless both versions are valid, so a malformed pin is never silently satisfied.

Regression tests added for each; `sfn fmt --check` / `sfn check` clean and `make compile` self-hosts.

---

Issue #1514, opened 2026-06-22
perf(llvm): emit arena mark/rewind around non-escaping generic-loop bodies


mcereal:
## Goal

Reclaim per-iteration non-escaping arena allocations by emitting `arena_mark`/`arena_rewind` around generic-`loop` bodies, so a hot allocation loop stops growing RSS linearly.

## Background

0.7.0 baseline (`docs/baselines/runtime-0.7.0-darwin-arm64.csv`): `benchmarks/runtime/arena_alloc.sfn` allocates 30M non-escaping `Cell` structs in a `loop { … }` and ends at **691 MB RSS = 1.01× the raw allocation size** — every iteration's struct is retained. The reclamation primitive already exists and is in the pinned seed (verified: `sfn_arena_sfn_mark`/`sfn_arena_sfn_rewind`, `runtime/sfn/memory/arena.sfn:620,656`, registered as intrinsics in `llvm/runtime_helpers.sfn:204-210`). The non-escape signal already exists too: locals start `allocation_kind="arena"` (`core_scopes.sfn:165`) and escapees are promoted to `rc` by `promote_escaping_local` (`core_scopes.sfn:332`); the survivors still `arena` + not `consumed` are exactly the non-escaping heap locals. The seam is `emit_scope_drops` at loop close (`instructions_loops.sfn:345`).

This issue **folds the soundness-predicate design** (originally a separate spike) into the implementation, per grooming decision.

## Strategy (decided: A — scope-rewind)

Emit `mark` at loop-body entry and `rewind` at iteration close. Chosen over scalar-replacement/stack-alloca (strategy B) because it reuses the shipped intrinsic with zero runtime/seed work and targets all per-iteration arena allocations. B is deferred post-1.0.

## Soundness predicate (must gate emission)

Wrap a loop body in mark/rewind ONLY when **all** hold:
- Every body-scope arena heap local is non-escaping: still `allocation_kind=="arena"`, not promoted to `rc`, not `consumed`.
- **No arena value allocated in the body is stored into / assigned to a binding whose scope is an ancestor of the loop body** (loop-carried escape — e.g. `let mut last; loop { last = Cell{...}; }`). This is the one landmine current promotion does NOT catch; rewinding here would free a value read next iteration. Detect via `body_result.locals` + mutation/assignment records at the seam; if any loop-carried arena store is present, **suppress rewind**.

## Scope

**In:**
[...]

Issue #1503, opened 2026-06-22
Epic: bench: native `sfn bench` command + `sfn/bench` harness


mcereal:
## Goal

Replace the bash benchmark scripts (`scripts/bench_compile.sh`, `scripts/bench_runtime.sh`) with a native, compiler-owned **`sfn bench`** command plus an in-language **`sfn/bench`** harness capsule — covering the compiler, the runtime, *and* user programs, with stable JSON output for agents.

Completes the compiler-owned-bench half of epic #513 Phase 3 (the determinism half already shipped as `sfn build --check-determinism`) and de-dups the bench slice of #1155.

## Design

The shape mirrors the existing `test` split — **CLI driver + stdlib capsule**:

| Layer | `test` precedent | `bench` analog | Owns |
|---|---|---|---|
| **CLI driver** (compiler orchestration) | `sfn test` (`cli_main.sfn`) | **`sfn bench`** | discovery, build + isolation, run N×, sample wall + peak RSS, aggregate min/median/stddev, budgets, table / CSV / JSON |
| **stdlib capsule** (in-language harness) | `sfn/test` (`capsules/sfn/test/`) | **`sfn/bench`** | `benchmark(name, fn)` API, iteration auto-calibration, `keep()`/black-box to defeat DCE, structured result emit |

**Three benchmark domains, one command:**
1. **compiler** — per-module compile time + memory (replaces `bench_compile.sh`); driver-only.
2. **runtime** — the bundled `benchmarks/runtime/*` workloads (replaces `bench_runtime.sh`); migrated onto `sfn/bench`.
3. **user** — a user's own `*_bench.sfn` via `sfn/bench`; same path as runtime.

**Two wins the native driver buys:**
- Drops the GNU `/usr/bin/time -v` dependency (absent on stock macOS, flaky in CI) — the driver samples child peak RSS itself via `getrusage`/`wait4`.
- Replaces the ad-hoc `RESULT <name> <ms> <ops>` stdout scrape with a structured driver↔capsule handshake (as `sfn test` ↔ `sfn/test`).

**Agent surface:** a versioned `sailfin.bench/v1` JSON envelope (mirroring `sailfin check --json`) + a `sailfin_bench` MCP tool.

## Sub-issues (DAG: A → {B, C} → D → E → F; C runs parallel to A)

- [ ] #1520 — **A** feat(runtime): subprocess rusage / peak-RSS capture *(foundation)*
[...]

PR #2135, opened 2026-07-11
feat(bench): sfn bench <path> runtime runner + migrate workloads onto sfn/bench (SFN-63)

mcereal:
## Summary

Adds the `sfn bench [<path>...]` runtime-execution runner and migrates the
bundled `benchmarks/runtime` workloads onto the `sfn/bench` harness (SFN-62),
retiring `scripts/bench_runtime.sh`. The runner discovers `*_bench.sfn`
workloads, builds each once, discards a warm-up, times K iterations via the
metered subprocess (`process.run_capture_metered`), parses the `bench-record/1`
line for inner-loop time + op count, samples peak RSS from rusage, aggregates
min/median inner-ms + peak RSS, checks budgets, and exits non-zero on any
build/run failure or budget violation. The table, CSV schema, and exit codes
match the retired bash script, so the `perf-history.yml` consumers are
unaffected.

Fixes SFN-63

## Changes

- **driver (`compiler/src/cli/commands/bench.sfn`)** — new default runtime mode
  (`_run_runtime_mode`); `--compiler` still selects the SFN-61 emit benchmark.
  Runtime flags: `--iterations` (5), `--warmup` (1), `--filter GLOB`, `--top`,
  `--csv`, `--budget-time` (median inner-ms), `--budget-mem` (peak RSS),
  `--work-dir` (default `build/bench-exec`), variadic `path` positional
  (default `benchmarks/runtime`). Discovery, glob filter, record parsing,
  min/median aggregation, table/summary/CSV/top-N, exit codes.
- **`compiler/src/cli/main.sfn`** — root help line for runtime-mode usage.
- **workloads (`benchmarks/runtime/*_bench.sfn`)** — all 10 renamed to
  `*_bench.sfn` and migrated onto `import { benchmark_fixed, keep } from
  "sfn/bench"`; the ad-hoc `RESULT` print is gone. The two arena-struct
  workloads batch their allocation in a directly-timed inner loop so the
  per-iteration arena rewind (#1514/#1515) keeps peak RSS flat.
- **`Makefile`** — `bench-runtime` is now one line: `sfn bench benchmarks/runtime`.
- **`scripts/bench_runtime.sh`** — deleted; comment references in
  `benchmark.yml` / `perf-history.yml` / `scripts/perf_history.sh` updated.
- **docs** — `benchmarks/runtime/README.md` rewritten; `docs/perf/*` and
  `docs/status.md` updated for the runner + record format + op-count methodology.
- **tests** — new `compiler/tests/e2e/bench_runtime_runner_test.sfn` (table,
  CSV schema, budget exit-2, empty-path exit-1); fixed the now-stale
  "missing --compiler exits 2" case in `bench_compiler_test.sfn`.

## Validation

- [x] `sfn fmt --check` clean on every touched `.sfn` file
- [x] `sfn check` (fast static analysis) passes
- [x] `make compile` — compiler self-hosts
- [ ] `make check` — full suite + seedcheck (running locally; CI also gates)
- [x] Regression coverage added (`bench_runtime_runner_test.sfn`)
- [x] `build/bin/sfn bench benchmarks/runtime` runs all 10 workloads, exit 0,
      table + CSV; `--budget-mem 1` exits 2; unknown path exits 1

## Capability envelope

Unchanged: the `bench` runtime mode uses only `io` + `clock` (subprocess build,
metered run, `fs` reads/writes, `uname` platform probe), the same effects the
existing `--compiler` mode already declared.

## Notes for reviewers

- **Compiler bug worked around (SFN-215):** `module_name_from_path` roots a
  module slug at any `runtime/` path segment, so a workload built in place from
  `benchmarks/runtime/` mis-slugs as a `runtime/` capsule module and fails to
  link its `sfn/bench` dependency. The runner stages each workload into a
  non-`runtime` `--work-dir` (default `build/bench-exec`, kept at repo depth 2
  so a workload's `../../runtime` relative import still resolves) before
  building. The staging comment cites SFN-215 as its removal condition.
- **Op-count methodology:** `benchmark_fixed(name, count, body)` reports `count`
  as the op count. Batch/round-unit workloads now report batch counts
  (`array_growth` 12,000, `array_map_filter` 8,000, `arena_reset_churn` 12,000,
  `exception_throw` 150,000, `arena_alloc`/`arena_alloc_for_range` 30,000)
  rather than raw element counts. The tracked metric stays median inner-ms +
  peak RSS; the CSV schema is unchanged.
- **Peak RSS:** the harness runs the whole timed window in one arena scope, so
  array/string workloads' peak RSS reflects the memory they allocate over the
  window (a real measurement of that path, documented in the README). The
  arena-struct workloads stay flat via the batched inner-loop rewind.

https://claude.ai/code/session_0145TcRHPW6cyS8D78CZmTNW

---
_Generated by [Claude Code](https://claude.ai/code/session_0145TcRHPW6cyS8D78CZmTNW)_


mcereal:
Thanks for the review — addressed in `34aeb1b`:

- **Conditional staging** (bench.sfn:716): good catch. The runner now stages into `work_dir` **only** when the path contains a `runtime/` segment (the SFN-215 case); otherwise it builds `src` in place, so relative imports resolve and `sfn bench <file>` is a proper drop-in runner. Verified with a non-`runtime` workload that imports a `./sibling` — it builds in place (no staged copy) and links.
- **ops/ms rounding** (bench.sfn:642): now rounds half-up to match the retired script's `printf "%.1f"`.
- **TMPDIR joins** (io_throughput_bench.sfn:22, bench_runtime_runner_test.sfn:58): guarded with a trailing-slash check.
- **Split inline-code span** (runtime-performance.md:32): record format is on one line now.

Not changed — **Darwin peak-RSS normalization** (bench.sfn:752): `ru_maxrss` is KiB on Linux and bytes on Darwin, and normalization is deferred at the primitive layer under epic #1503 (documented in `runtime/sfn/process.sfn`, gated in `process_run_capture_metered_test.sfn`, and matched by the already-merged `sfn bench --compiler` mode which reads the same primitive). Normalizing in this one consumer would diverge from the primitive and the sibling compiler mode, so I left it to the tracked #1503 fix and added a comment noting the peak-RSS column and `--budget-mem` are Linux/WSL-oriented today.


---

